PR #54, opened 2013-10-11
Ukraine local flavor

bacher09:
Local flavor for Ukraine.


benkonrath:
@bacher09 I'm closing this PR since it hasn't been updated in a while. If you're still interested in contributing this localflavor, please file a new PR with this change so that a clean review can be done. Thanks.

PR #63, opened 2013-12-09
Add support for Australian Business Numbers

danni:


jscn:
I'd like to see this merged. Is it okay to rebase onto master and push to this PR, or should I create a new PR?

danni:
@jscn go for it

jscn:
@jezdez @erikr I think I've addressed both your concerns, care to take a look?

jscn:
ping? @erikr @jezdez

benkonrath:
erikr and jezdez haven't spent much time on this project recently. I can take over this PR review but I'm a little pressed for time right now. I might able to spend some time next week.

jscn:
No worries, thanks for the update.

benkonrath:
A quick update: I'm at DjangoCon EU and plan to look at this during the sprints on Saturday and Sunday (if I don't find time before then).

localflavor/au/forms.py
```diff
@@ -63,3 +65,22 @@ class AUStateSelect(Select):
...
+    .. versionadded:: 1.3
+    """
+
+    default_validators = [AUBusinessNumberFieldValidator()]
+
+    def clean(self, value):
+        """
+        Return a cleaned ABN.
+        """
+        value = super(AUBusinessNumberField, self).clean(value)\
```
benkonrath: There should be a space before the '\'. 

You don't actually need to put the '.replace(...)' on a separate line because we're using 120 character lines in this project. That said, we're not enforcing a specific style yet so you can keep the shorter lines if that's your preference.

localflavor/au/validators.py
```diff
@@ -0,0 +1,56 @@
...
+        Return whether the given value is a valid ABN.
+
+        See https://en.wikipedia.org/wiki/Australian_Business_Number
+        for a description of the validation algorithm.
+        """
+        # ABNs must be 11 digits, ignoring whitespace.
+        try:
+            digits = [int(i) for i in list(value.replace(' ', ''))]
+        except ValueError:
+            return False
```
benkonrath: You don't need the try / except because the regex already enforces the digit only constraint.

localflavor/au/models.py
```diff
@@ -52,3 +53,26 @@ def formfield(self, **kwargs):
...
+    .. versionadded:: 1.3
+    """
+
+    description = _("Australian Business Number")
+
+    validators = [AUBusinessNumberFieldValidator()]
+
+    def __init__(self, *args, **kwargs):
+        # ABN is 11 digits long, but we allow to enter it using spaces
+        kwargs['max_length'] = 14
```
benkonrath: I don't think you should store the value with spaces in the database. With the current design, '5300 4085 616' and '53004085616' would not be considered the same value even though it's the same ABN. This means a unique constraint on the field would miss some cases.

I hit a similar problem with spaces when I developed the `IBANField`. You can remove the spaces in   `to_python()` of the model field like this:
https://github.com/django/django-localflavor/blob/master/localflavor/generic/models.py#L52

Doing this is not without issues though. If a users enters '5300 4085 616', they will see  '53004085616' after they save the form. It is also possible to re-add the spaces in the value on the form in `prepare_value`. Here's an example:
https://github.com/django/django-localflavor/blob/master/localflavor/generic/forms.py#L97

Of course this also has a problem: if a user enters ''53004085616', they will see '5300 4085 616' when the form is saved.

In the end I'm not sure how an ABN is normally displayed so I'd don't know if adding spaces automatically like the IBANField is best approach. You're free to decide what makes sense for this. Let me know if you agree with saving only the digits in the database.

localflavor/au/validators.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+    Validation for Australian Business Numbers.
+
+    .. versionadded:: 1.3
+    """
+
+    error_message = _('Enter a valid ABN.')
+
+    def __init__(self):
+        eleven_digits_with_whitespace = '^(\s*\d\s*){11}$'
```
benkonrath: This should be updated to only check for digits since the spaces can be removed elsewhere.

localflavor/au/validators.py
```diff
@@ -0,0 +1,56 @@
...
+        super(AUBusinessNumberFieldValidator, self).__init__(
+            regex=eleven_digits_with_whitespace, message=self.error_message)
+
+    def _is_valid(self, value):
+        """
+        Return whether the given value is a valid ABN.
+
+        See https://en.wikipedia.org/wiki/Australian_Business_Number
+        for a description of the validation algorithm.
```
benkonrath: A link to the validation algorithm would be nice. I found this which seems useful:
http://www.clearwater.com.au/code

tests/test_au/tests.py
```diff
@@ -106,3 +118,86 @@ def test_AUPhoneNumberField(self):
...
+    def test_no_error_for_a_valid_abn(self):
+        """Test a valid ABN does not cause an error."""
+        valid_abn = '53004085616'
+
+        validator = AUBusinessNumberFieldValidator()
+
+        try:
+            validator(valid_abn)
+        except ValidationError:
+            self.fail('Validation unexpectedly failed.')
```
benkonrath: You don't need the try / except. You can just run the validator. If it fails, the tests will break which is what we want.

tests/test_au/tests.py
```diff
@@ -106,3 +118,86 @@ def test_AUPhoneNumberField(self):
...
+    def test_allows_whitespace(self):
+        """Test an ABN can be valid when it contains whitespace."""
+        valid_abn = '5300 4085 616'
+
+        validator = AUBusinessNumberFieldValidator()
+
+        try:
+            validator(valid_abn)
+        except ValidationError:
+            self.fail('Validation unexpectedly failed.')
```
benkonrath: Again, you don't need the try / except.

benkonrath:
@jscn Sorry for the delay in reviewing this. As I mentioned, I wanted to do this during the sprints at DjangoCon EU but there were more people at the sprints that wanted to help out with localflavor than I expected. I've added my comments now and I'm looking forward to your thoughts. Thanks.

jscn:
@benkonrath I think I've fixed all the problems you identified, now. I'd really like to get this in for the 1.3 release, so let me know as soon as you can if there are any other changes you'd like to see.

jscn:
:/ apparently `isort` and I disagree about correct ordering of imports.

benkonrath:
@jscn I'm just back online from a short vacation. Thanks for reworking this - it looks good. Just a couple of housekeeping items left:

* Add an entry to the docs/changelog.rst describing the change.
* Add an entry for your name and @danni's name to the docs/authors.rst file if it's not already there.
* Rebase and squash the commits into 1 commit.

I'd like to get this into 1.3 as well - I can delay by a day or two if you need a bit of time to sort this out. Thanks again for your work on this feature.

jscn:
@benkonrath Thanks for your help getting this in shape, I think it's all ready to go now, but let me know if there are further changes needed.

benkonrath:
Thank you too. :)

PR #137, opened 2014-12-10
Adding US State Level localflavors

igorzelenberg:
Adding state level localflavors: adding New York State Counties first.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -73,6 +73,7 @@ validate Finnish social security numbers.
    * :doc:`localflavor/sk`
    * :doc:`localflavor/tr`
    * :doc:`localflavor/us`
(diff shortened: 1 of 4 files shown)


ambsw-technology:
I actually found this issue when looking for any discussion on counties.

I can imagine a state-level field working for state governments and highly regulated industries like insurance, but not much else.  That isn't to say it shouldn't be merged (in fact the county data can be leveraged in a more widely-used model), but doesn't make a strong case for it either.

I suspect most people need a national county field that cross-filters on state.  At minimum, some kind of back-end validation could be added, but a useful field would really require some form magic.  I have a feeling that kind of functionality is out-of-scope for this package so it's probably a lost cause (though I too want/need it).

PR #142, opened 2015-02-16
Proposal: Use libphonenumber for phone number validation

kane-c:
The phone number validation rules are quite simple at the moment. For example, Australian numbers must be 10 digits.

This PR uses the excellent [port of `libphonenumber`](https://github.com/daviddrysdale/python-phonenumbers) to have thorough phone number validation for just about every location.

I've included the option to allow/disallow international numbers as part of the validation, and to be able to choose which number types to allow (e.g. mobile only, land line only, etc.)

I've only included the new validation in the Australian phone number field for now however if this is accepted, we can easily add it to other countries too.

### DIFF
--- a/localflavor/au/forms.py
+++ b/localflavor/au/forms.py
@@ -6,11 +6,10 @@ from __future__ import absolute_import, unicode_literals
 
 import re
 
-from django.core.validators import EMPTY_VALUES
-from django.forms import ValidationError
-from django.forms.fields import CharField, RegexField, Select
-from django.utils.encoding import smart_text
+from django.forms.fields import RegexField, Select
 from django.utils.translation import ugettext_lazy as _
+
+from localflavor.generic.forms import PhoneNumberField
 
 from .au_states import STATE_CHOICES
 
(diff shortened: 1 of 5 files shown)


benkonrath:
I like this. One problem off the top of my head is that phone numbers already stored for the current phone number fields may not pass the libphonenumber validation as your changes to the test_AUPhoneNumberField method demonstrates. It could happen that previously valid data would become invalid by just updating to a new version of localflavor.

What about only making a generic PhoneNumberField without changing the implementation of the current phone number fields. The current fields could be deprecated and eventually removed.

Having a model field would also be useful.

Some other points to consider: I'm not sure if localflavor has an official deprecation policy yet and I don't know how other localflavor devs feel about adding a dependency like libphonenumber. Any opinions here are welcome.

benkonrath:
Since there hasn't been any objections to the `phonenumbers` dependency, I'll assume it's ok.

@kane-c Sorry for the delay in getting to this. Are still interested in working on this PR? If so, can you please rebase it and update to the latest version of `phonenumbers`? I can do a detailed review once this is done. Thanks.

kane-c:
Hi @benkonrath. I just needed this update for a project I was working on and thought it might be useful to include upstream. If there's interest, I can update it though.

benkonrath:
@kane-c Yes, I'm still interested. I'm on vacation for the next week so can't do a proper review until I'm back.

I skimmed the PR. From a hight level, I would prefer to copy the style of the `IBANField` and have an `include_countries` argument rather than making a sub-class for each country. New projects would use the `PhoneNumberField` directly which could support multiple countries with just the one field (if required). It would still be possible to have a separate phone number field for each country if this is more suitable for the requirements of the project. The existing phone number fields could be implemented with the new `PhoneNumberField` until we have time to gracefully deprecate them.

I'm curious to hear your thoughts and the opinion of the other collaborators. I'll check in again in a week or so. Thanks!

benkonrath:
@kane-c Just checking in to see if you're still interested in working on this.

kane-c:
@benkonrath Yep, still interested. I'll rebase and work through your recommendations shortly.

benkonrath:
After looking at this again, I think it's better to recommend that you use the django-phonenumber-field which already has a working phone number field based on libphonenumer and is actively maintained.
https://github.com/stefanfoulis/django-phonenumber-field

There's a proposal to deprecate the phone number fields in localflavor and recommend people use django-phonenumber-field instead (see #223). I think is a good way forward for dealing with the incorrect phone number validations in localflavor without bringing in a large dependency (even if it is optional) and without reinventing the wheel.

Thanks for your work on this pull request.

kane-c:
Ah great, I didn't know about that project. I'll use it in the future. Thanks.

PR #143, opened 2015-03-04
Update django.po

gabejackson:
This error message is used for austrian, swiss and norwegian forms. It shouldn't contain a specific country.


erikr:
Thanks for your patch, this seems entirely sensible! However, all translation management is [done through Transifex](https://www.transifex.com/projects/p/django-localflavor/). Would you mind submitting your change there?

claudep:
FTR, I've just fixed those strings in Transifex.

jezdez:
Thanks @claudep!

PR #112, opened 2014-08-08
Chile local flavor: New form select field: CLCommuneSelect

imunizaga:
Included a list of commune as choices in localflavor/cl/cl_communes
Included a select field in localflavor/cl/forms.py

Also included a dict with the name of the communes mapping to the region
they belong to in localflavor/cl/cl_regions, in case someone wants to
validate the correspondence of the commune in the region.

Finnaly, changed the localflavor/cl/forms.py file to fit every line in
80 characters

PR #144, opened 2015-03-04
Chile local flavor: Add form select field: CLCommuneSelect

thagat:
Include a list of commune as choices in localflavor/cl/cl_communes
Include a commune select field in localflavor/cl/forms.py

Also include a dict with the name of the communes mapping to the region
they belong to in localflavor/cl/cl_regions, in case someone wants to
validate the correspondence of the commune in the region.

Include a test for the new communes CLCommuneSelect field


glarrain:
What's the convention for form choices regarding the first item in each tuple? I've usually seen things like this: `('monte_patria', 'Monte Patria')` instead of `('Monte Patria', 'Monte Patria')`. Perhaps only-lowercase alphanumeric characters (plus underscore) would be better.

nagisa:
Is this the retry of https://github.com/django/django-localflavor/pull/112? Can that other one be closed then?

PR #146, opened 2015-03-10
Strip field in order to avoid int conversion crash

viorels:
If the CIF ended with a space the validation crashed with this error: invalid literal for int() with base 10: ''

### DIFF
--- a/localflavor/ro/forms.py
+++ b/localflavor/ro/forms.py
@@ -36,7 +36,7 @@ class ROCIFField(RegexField):
         """
         CIF validation
         """
-        value = super(ROCIFField, self).clean(value)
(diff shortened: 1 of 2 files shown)


avelino:
crash tests, pls fixed!

viorels:
Don't know why all tests are failing but it doesn't seem to be related to my patch. Any idea?

nagisa:
Yes, this doesn’t seem to be relevant. It looks like the lint was updated recently with a new heuristics, that our current code base doesn’t pass.

Patch LGTM, would merge if I was a maintainer of the Romanian local flavour.

claudep:
Merged in cbff14614c, thanks.

Issue #150, opened 2015-04-15
local-flavor NL: add model fields, renaming, testing, deprecating.


jieter:
[As mentioned in this PR comment](https://github.com/django/django-localflavor/pull/148#issuecomment-92922485) I'm about to do some work on the Dutch local-flavor fields. While working on reorganizing and adding the model fields and tests, a couple of questions rise:
- [x] Since everyone is now required to use IBAN, there is no use for `NLBankAccountNumber*` anymore. I suggest removing(or at least deprecating) it.
- [x] The term 'Sofinummer' is succeeded in common use by the term 'Burger Service Nummer', I suggest renaming `SoFi` to BSN. (There is an [edge case](http://nl.wikipedia.org/wiki/Sofinummer#Wanneer_wordt_een_sofinummer_toegekend.3F))
- [ ] It seems to be common to call model fields `FooField` and form fields `FooFormField`. This is not pattern used for the current fields. Should we rename the form fields?

cc @benkonrath

benkonrath:
Point 1: Some users may actually still want to store the old Dutch bank numbers but I don't know for sure. I suppose the question is really about whether localflavor should keep historical fields or not. If we were to remove the `NLBankAccountNumber*` fields, they would need to be depreciated first and I think it would be better to file a separate issue to solve this generally for all the SEPA bank fields in localflavor. 

Point 2 and Point 3: You would need to provide South and native Django migrations for anything you want to rename. Although I don't know off hand what the localflavor policy for renaming fields is.

@jezdez @erikr @claudep Are any of you guys able to provide more information here? Thanks.

erikr:
On point 1: I think we should deprecate all local bank account fields, in countries where IBAN has become mandatory. By mandatory I mean that in the Netherlands, I can no longer enter an old form account number in a bank transfer, and the conversion service has shut down. So there is no possible use for an old form account number anymore. I believe this is part of SEPA, and therefore required in all SEPA countries. For non-SEPA IBAN countries we should then keep the existing field. Also, special territories of some countries are [excluded from SEPA](https://en.wikipedia.org/wiki/Single_Euro_Payments_Area#Coverage), like Greenland or Sint Maarten, so if they still use the Danish/Dutch old formats, we need to keep the fields of that country.

[...]

PR #152, opened 2015-05-09
Clean up Dutch local flavor.

jieter:
This is my initial attempt on #150. As of now, nothing is (or at least should be) renamed or deprecated. This is what I did:

 - Move validators to validators.py
 - Add model fields for existing form fields
 - Add form fields for existing model fields
 - Move tests to `tests/test_nl/`
 - Increase test coverage.
 - Tried to make the error messages a bit more consistent.

## Known issues / TODOs:
 - Add docstrings
 - Tests fail with django 1.5: `RegExvalidator` somehow fails with message `Enter a valid value.` and not `Enter a valid zip code.`. Any pointers to fix this?

Please review!


jieter:
@erikr @benkonrath ?

benkonrath:
@jieter Sorry, this fell off my radar. I'll try to look at it later today.

localflavor/nl/validators.py
```diff
@@ -0,0 +1,127 @@
...
+
+        if int(value[:4]) < 1000:
+            raise ValidationError(self.error_message)
+
+
+class NLSoFiNumberFieldValidator(RegexValidator):
+    """
+    Validation for Dutch Sofinummers.
+
+    TODO: consider renaming to BSN
```
benkonrath: Yeah, I agree that it should be changed to BSN but let's do this as separate issue.

localflavor/nl/models.py
```diff
@@ -1,62 +1,88 @@
...
+        defaults = {'form_class': forms.NLSoFiNumberField}
+        defaults.update(kwargs)
+        return super(NLSoFiNumberField, self).formfield(**defaults)
+
+
+class NLPhoneNumberField(models.CharField):
+    """
+    Dutch phone number model field
+    """
+    description = _('Dutch phone number')
```
benkonrath: The comment is redundant with the description right there.

localflavor/nl/models.py
```diff
@@ -1,62 +1,88 @@
...
+        return '%s %s' % (value[:4], value[4:])
+
+    def formfield(self, **kwargs):
+        defaults = {'form_class': forms.NLZipCodeField}
+        defaults.update(kwargs)
+        return super(NLZipCodeField, self).formfield(**defaults)
+
+
+class NLProvinceField(models.CharField):
+    description = _('Dutch social security number (SoFi)')
```
benkonrath: This is the NLProvinceField, not the SoFi field. It's probably just a cut 'n paste error. :)

benkonrath:
Besides those small points, it looks good. If you can fix these points, figure why the tests fail on 1.5 and get the docstrings written, then we can try to get it in the next release which is due shortly. Otherwise we'll shoot for the release after next when 1.5 support is dropped as per #170.

localflavor/nl/models.py
```diff
@@ -1,62 +1,88 @@
...
+    def __init__(self, *args, **kwargs):
+        kwargs['max_length'] = 7
+        super(NLZipCodeField, self).__init__(*args, **kwargs)
 
-    def __init__(self, regex=None, message=None, code=None):
-        super(NLBankAccountNumberFieldValidator, self).__init__(regex='^[0-9]+$',
-                                                                message=self.default_error_messages['invalid'])
-        self.no_leading_zeros_regex = re.compile('[1-9]+')
+    def to_python(self, value):
+        return '%s %s' % (value[:4], value[4:])
```
benkonrath: I found a problem here. This is adding an extra space when the model field is used with the form field. For example, if I enter '1000AA', I will get the error message: 'Ensure this value has at most 7 characters (it has 8).' because the form adds a space and then this adds a space. You'll need to check if the space is already there before adding it. You should also add a test for this. Thanks.

jieter:
Thanks for the feedback, I'll come back to this, hopefully next week. The delay pushed it somewhat down on my todo-list...

benkonrath:
Understandable. I'll keep my eyes peeled for your update. Thanks.

localflavor/nl/models.py
jieter: This is in the code multiple times, should I remove them all? Are they not useful while generating the docs?

benkonrath: Yes, good point, they are used for the docs. Please ignore my comment.

localflavor/nl/models.py
jieter: Shouldn't we just save it in the database without space and present it with a space?

jieter:
Just pushed some improvements. I'm not going to invest time in adding 1.5 support if it's about to be removed anyway, so I'll guess we have to wait until after the next release then?

localflavor/nl/models.py
jieter: Hmm, this seems to be fixed by the last commit I pushed.

benkonrath:
I think it makes sense to wait until localflavor 1.2 is released so that we can skip trying to fix the errors with Django 1.5. Although I'm not sure when that's going to happen.

The bug in the NL postal code model field in fact fixed but it would be nice to rework that bit of code to be more like the IBAN model. 

You should make sure you get 100% test coverage. There's a bit of the NL postal code model field that isn't tested and the bank account model field isn't tested as well. I know that we want to  deprecate and drop the EU bank account fields but we should keep the tests in until the fields are removed.

Thanks for your work on this.

jieter:
- Just noticed that I added forms.NLBankaccountField in this branch, so I removed it. We should not add new code to deprecate it immediately.
- re-added the test for the bank account model field, you're right about that.
- test coverage is now at 100%.
- the Zip code form field needs the six.string_types check because apparently an array might be passed.

benkonrath:
Nice catch on the forms.NLBankaccountField - I did see it but I just assumed it was already there. Nice work here - l'll merge this after localflavor 1.2 is released.

benkonrath:
I forgot to mention that you should rebase this and squash all your commits into one. Thanks.

jieter:
yes, will do. And what about the `.. versionadded:: TODO` markers?

benkonrath:
We'll have to wait to see what the version will be after 1.2. You can wait to squash everything once the `versionadded` comments are in.

jieter:
Will squash again then.

jieter:
All tests pass now, should I extend the docstrings with `..versionadded::1.3` now?

benkonrath:
Yeah, `..versionadded:: 1.3` would be great for any new classes.

Can you also add an entry to the changelog for this PR? The moved validator classes should be specifically mentioned as a breaking change so users know they will need to adjust their imports. I want to setup a deprecation policy for localflavor but we can sort that out before the next release.

The final (small) change it to reset the commit date. Simply running `git commit --amend --reset-author` on your branch should take care of it.

jieter:
This should do it.

benkonrath:
@jieter Thanks for your work on this!

jieter:
@benkonrath any idea on when a version containing this PR is going to be released to pypi?

benkonrath:
@jieter Within the next week or two I hope. See my comment about this in #205.

jieter:
I see, thanks for the update

PR #154, opened 2015-06-02
fix ch zip

camillobruni:
According to <http://en.wikipedia.org/wiki/List_of_postal_codes_of_Switzerland> and
<http://en.wikipedia.org/wiki/Postal_codes_in_Switzerland_and_Liechtenstein> the Swiss
zip code has a range from 1000-9000 similar to the belgian ones. The current
implementation wrongly accepts 0000 as a valid zip code.

### DIFF
--- a/localflavor/ch/forms.py
+++ b/localflavor/ch/forms.py
@@ -14,24 +14,29 @@ from django.utils.translation import ugettext_lazy as _
 
 from .ch_states import STATE_CHOICES
 
+zip_re = re.compile(r'^[1-9]\d{3}$')
 
 id_re = re.compile(
     r"^(?P<idnumber>\w{8})(?P<pos9>(\d{1}|<))(?P<checksum>\d{1})$")
+
 phone_digits_re = re.compile(r'^0([1-9]{1})\d{8}$')
 
 
 class CHZipCodeField(RegexField):
     """
     A form field that validates input as a Swiss zip code. Valid codes
-    consist of four digits.
+    consist of four digits ranging from 1XXX to 9XXX.
+
+    See:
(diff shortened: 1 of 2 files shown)


localflavor/ch/forms.py
```diff
@@ -14,24 +14,46 @@
...
+    - 5xxx - Region Aarau
+    - 6xxx - Region Central Switzerland (Innerschweiz), Tessin
+    - 7xxx - Region Graubuenden
+    - 8xxx - Region Zurich
+    - 9xxx - Region Eastern Switzerland (Ostschweiz)
+
+    The range 9480 to 9499 is assigned to zip codes from Liechtenstein.
+
+    See:
+    http://en.wikipedia.org/wiki/Postal_codes_in_Switzerland_and_Liechtenstein
```
claudep: I'm not sure this detailed explanation is useful here. I would simply add the range and the wikipedia link.

localflavor/ch/forms.py
```diff
@@ -14,24 +14,46 @@
...
     """
+
     default_error_messages = {
-        'invalid': _('Enter a zip code in the format XXXX.'),
+        'invalid': _(
+            'Enter a valid postal code in the range and format 1XXX - 9XXX.'),
```
claudep: Please use a single line here, we are not mad at the 80 char limit.

camillobruni: ok ;) against my usual behavior I stuck with what I saw

claudep:
Thanks, looks good now. Could you please merge your commits?

camillobruni:
there you go, thanks for the quick feedback

claudep:
Test failures are unrelated to this patch. Thanks for your work.

PR #155, opened 2015-06-02
Adding support for Swiss SSN

camillobruni:
See 
    http://www.bsv.admin.ch/themen/ahv/00011/02185/
    http://de.wikipedia.org/wiki/Sozialversicherungsnummer#Versichertennummer

### DIFF
--- a/localflavor/ch/forms.py
+++ b/localflavor/ch/forms.py
@@ -7,19 +7,20 @@ from __future__ import absolute_import, unicode_literals
 import re
 
 from django.core.validators import EMPTY_VALUES
+from django import forms
 from django.forms import ValidationError
 from django.forms.fields import Field, RegexField, Select
 from django.utils.encoding import smart_text
 from django.utils.translation import ugettext_lazy as _
 
 from .ch_states import STATE_CHOICES
+from ..generic import validators
 
 zip_re = re.compile(r'^[1-9]\d{3}$')
-
 id_re = re.compile(
     r"^(?P<idnumber>\w{8})(?P<pos9>(\d{1}|<))(?P<checksum>\d{1})$")
-
 phone_digits_re = re.compile(r'^0([1-9]{1})\d{8}$')
+ssn_re = re.compile(r'^756.\d{4}\.\d{4}\.\d{2}$')
 
 
 class CHZipCodeField(RegexField):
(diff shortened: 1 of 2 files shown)


localflavor/ch/forms.py
```diff
@@ -130,3 +132,54 @@ def clean(self, value):
             raise ValidationError(self.error_messages['invalid'])
 
         return '%s%s%s' % (idnumber, pos9, checksum)
+
+
+class CHSocialSecurityNumberField(forms.CharField):
```
claudep: Just a question: did you try to use `RegexField` like for `CHZipCodeField`?

camillobruni: I thought about it but then based it on the the US-version (which is a `CharField` AFAIK).
Probably a `RegexField` plus a separate EAN-13 validation would do the trick.

claudep:
Very good idea to add a new common ean validator. But I'd like to see that as a separate pull request, if possible (mentioning that you'd like to use it at least for Swiss AHV).

camillobruni:
Can do :+1:

claudep:
Now this one can be rebased.

camillobruni:
Rebased onto the latest master, should be ready for merging.

localflavor/ch/forms.py
```diff
@@ -135,3 +136,31 @@ def clean(self, value):
...
+
+    See:
+    http://de.wikipedia.org/wiki/Sozialversicherungsnummer#Versichertennummer
+    """
+    default_error_messages = {
+        'invalid': _('Enter a valid Swiss Social Security number in 756.XXXX.XXXX.XX format.'),
+    }
+
+    def __init__(self, max_length=None, min_length=None, *args, **kwargs):
+        super(CHSocialSecurityNumberField, self).__init__(ssn_re, max_length, min_length, *args, **kwargs)
```
claudep: Would it be possible to pass the EAN validator here in the `validators` keyword argument of `__init__`? Then we may even get rid of the `clean` method (to be evaluated).

localflavor/ch/forms.py
```diff
@@ -7,19 +7,20 @@
 import re
 
 from django.core.validators import EMPTY_VALUES
+from django import forms
```
claudep: I think this import is not needed. `RegexField` is already imported individually just below.

localflavor/ch/forms.py
camillobruni: This has been my initial approach, but then you get duplicated error messages as all the validators are run together (maybe I missed something...). In this case the user experience is not that nice, since you would get a warning for instance about the failed EAN checksum validation, I'd prefer to just have the error message from `CHSocialSecurityNumberField`. Any suggestions?

localflavor/ch/forms.py
camillobruni: Indeed it is not, I can remove it. I did it to follow other Field implementations :)

localflavor/ch/forms.py
claudep: I see. But if we bypass the regex validation in clean, there is no sense to subclass `RegexField`. Subclassing `CharField` and passing validators in `__init__` should do the trick.

I would also add `code='invalid'` when raising `ValidationError` in `EANValidator`, and then we can also omit passing `message` to the validator.

camillobruni: seems reasonable, I'll try

claudep:
Camillo, I've amended your patch and created this PR: #158
Are you OK with that variant and keeping you as the commit author?

camillobruni:
oh, certainly fine with me :+1:

claudep:
Thanks, I'll wait for the CI to finish and commit that.

claudep:
Closed in favor of #158

PR #156, opened 2015-06-03
Add separate EAN validator,

camillobruni:
In prepration for new Swiss SSN / AHV Field

### DIFF
--- a/localflavor/generic/validators.py
+++ b/localflavor/generic/validators.py
@@ -1,11 +1,15 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+import re
+
 import string
 
 from django.core.exceptions import ValidationError, ImproperlyConfigured
 from django.utils.translation import ugettext_lazy as _
+
 from .countries.iso_3166 import ISO_3166_1_ALPHA2_COUNTRY_CODES
+from . import checksums
 
 # Dictionary of ISO country code to IBAN length.
 #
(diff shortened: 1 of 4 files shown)


localflavor/generic/validators.py
```diff
@@ -209,3 +212,23 @@ def __call__(self, value):
...
+
+class EANValidator(object):
+    """
+    A generic validator for EAN like codes with the last digit being the checksum.
+
+    http://en.wikipedia.org/wiki/International_Article_Number_(EAN)
+    """
+
+    def __init__(self, strip_nondigits=False, error_message=None):
+        self.error_message = error_message
```
claudep: Could we mimic the error message implementation of Django core validators? That is a default `message = _(...)` on the class, a `message` keyword argument and the `if message is not None:  self.message = message` sequence in `__init__`.

localflavor/generic/validators.py
```diff
@@ -1,11 +1,13 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
+import re
```
claudep: nitpick: blank line before, not after `import re`

tests/test_generic.py
```diff
@@ -337,3 +337,63 @@ def test_bic_model_field(self):
...
+            None,
+        ]
+        error_message = 'Invalid EAN Code'
+        invalid = {
+            '400.6381.3339.31',
+            '4006381333930',
+            '',
+            '0',
+            'DÉUTDEFF',
+        }
```
claudep: This set constructor syntax is not supported on Python 2.6 and localflavor is still supporting Python 2.6.

tests/test_generic.py
```diff
@@ -337,3 +337,63 @@ def test_bic_model_field(self):
...
+        ]
+        error_message = 'Invalid EAN Code'
+        invalid = {
+            '4006381333930',
+            '400-63-813-339-30',
+            '400 63 813 339 30',
+            '',
+            '0',
+            'DÉUTDEFF',
+        }
```
claudep: Same as above.

tests/test_generic.py
camillobruni: should've been a list, thanks

claudep:
Tests are now failing with Python 3, see Travis CI details.

PR #157, opened 2015-06-19
USSocialSecurityNumberField: proper check for promotional SSNs

ataylor32:


claudep:
Just a note that test failures are not related to this patch, they are fixed by 957bc3761f

benkonrath:
Thanks for the fix and sorry for the delay. Committed via f31007072edc82c7ebc6dacafa1f5242a40a11c7.

PR #158, opened 2015-06-24
Adding support for swiss SSN

claudep:


claudep:
Merged in 9abf18c36c8

PR #165, opened 2015-08-15
Change error string for ITSocialSecurityNumberField

zarelit:
Use the same terms as Agenzia delle Entrate


claudep:
Translations are handled through Transifex, so you should remove .po/.mo changes from your patch, please.

PR #167, opened 2015-08-15
Fix Italian Tax code error message

### LINKED ISSUE #164: localflavor-it: wrong error string in ITSocialSecurityNumberField
The original error string in it/forms.py is "Enter a valid Social Security number." that is literally translated in the Italian language file.

While this is good for the same string in another context (e.g. the Mexican forms.py) it has the wrong meaning if used in Italian applications; the error string should be something like "Enter a valid Fiscal Code" or "Inserisci un Codice Fiscale valido".

The string "Inserisci un numero di Assistenza Sociale valido" cannot be understood by Italian non English-speaking people.

yakky:
Fix #164

### DIFF
--- a/localflavor/it/forms.py
+++ b/localflavor/it/forms.py
@@ -60,8 +60,7 @@ class ITProvinceSelect(Select):
 
 class ITSocialSecurityNumberField(RegexField):
     """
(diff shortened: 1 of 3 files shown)


claudep:
Dup of #166. @zarelit Should we commit this one or your patch?

PR #169, opened 2015-09-02
Correct sepa countries list based on the latest data.

benkonrath:
It looks like Croatia was missed in the first version of this.

http://www.europeanpaymentscouncil.eu/index.cfm/knowledge-bank/epc-documents/epc-list-of-sepa-scheme-countries/


claudep:
Looks like Croatia was not missing but misplaced (maybe the list was ordered by country name at some point rather than by code name).

benkonrath:
Good catch. I'll fix this. Thanks.

benkonrath:
I fixed the order and removed the duplicate 'ES' entry for the Canary Islands.

claudep:
Thanks Ben.

PR #170, opened 2015-09-02
Drop support for Django 1.5 / 1.6 and Python 2.6

benkonrath:
Motivated by the unknown Django 1.5 test errors in #152, I propose we drop support for unsupported versions of Django (1.5 and 1.6) which means we can also drop Python 2.6 support. This would be for the next release of localflavor (1.2). I'm curious to hear what people think about this.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,5 +1,4 @@
 invoke==0.11.1
 coverage==3.7.1
 flake8==2.2.5
-django-discover-runner==1.0
 six==1.8.0
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -1,4 +1,4 @@
 invoke==0.11.1
 coverage==3.7.1
 flake8==2.2.5
-six==1.8.0
+six==1.10.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,14 +7,6 @@ env:
 matrix:
   include:
     - env: TOXENV=docs
-    - python: 2.6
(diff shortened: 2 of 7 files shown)


claudep:
My opinion is that we should release ASAP as I know about several people waiting for a release with recent fixes.
Then I'm OK to drop 1.5/1.6 for next release+1.

benkonrath:
That's ok by me. Any idea when the next release is coming?

claudep:
As soon as one of the releasers finds time to do it. Sorry, I cannot be more precise.

benkonrath:
No problem, this answers my question actually. I wasn't sure if you meant ASAP to mean in the next couple of days. Thanks.

jieter:
Would be nice if we could move on with this...

benkonrath:
I've just updated / rebased this PR. I'm just looking for a quick sanity check. Thanks.

tests/requirements.txt
```diff
@@ -1,5 +1,4 @@
 invoke==0.11.1
 coverage==3.7.1
 flake8==2.2.5
-django-discover-runner==1.0
 six==1.8.0
```
claudep: Maybe six>1.8.0 (or six==1.10.0)?

claudep: But it doesn't need to be in that PR, in fact.

benkonrath:
I included the updated six version in this PR. Thanks.

claudep:
Looks good, please push it!

PR #172, opened 2015-10-07
Added CNPJ and CPF validation formats

LuanP:
Changed the re.sub to avoid cleaning CNPJ and CPF numbers sent in the wrong format.

Removed errors_numbersonly since the regex will take care of validating and raising a ValidationError with the error_format message.

If the message of errors_numbersonly error must be kept, just let me know.

### DIFF
--- a/localflavor/br/forms.py
+++ b/localflavor/br/forms.py
@@ -20,6 +20,10 @@ except ImportError:
 from .br_states import STATE_CHOICES
 
 phone_digits_re = re.compile(r'^(\d{2})[-\.]?(\d{4,5})[-\.]?(\d{4})$')
+cpf_digits_re = re.compile(r'^(\d{3})\.(\d{3})\.(\d{3})-(\d{2})$')
+cnpj_digits_re = re.compile(
+    r'^(\d{2})[.-]?(\d{3})[.-]?(\d{3})/(\d{4})-(\d{2})$'
+)
 
 
 class BRZipCodeField(RegexField):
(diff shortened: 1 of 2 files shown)


LuanP:
I'd like to bring the validations out of the clean method. Allowing users to use the validation functions directly. Could I do that?

Thanks

claudep:
When reading the test updates, it's not obvious for me what you are trying to solve with this patch. If the only result is to have a different error message, it's not worth it I think. If your changes allow to catch more errors, please reflect that in the tests also.

LuanP:
I've added a CNPJ number to the tests:
https://github.com/LuanP/django-localflavor/blob/change_cpf_and_cnpj_regexp/tests/test_br.py#L34

This PR makes that invalid.

claudep:
Good, a similar test for CPF would be nice.

LuanP:
@claudep oh... thanks, I forgot that. u.u

Added in
https://github.com/LuanP/django-localflavor/blob/change_cpf_and_cnpj_regexp/tests/test_br.py#L54

LuanP:
@claudep Do you have any ideas about what happened with the builds py27-master and py34-master that raised an AssertionError in the Chilean RUT? Every other env has passed successfully.

Anything else I should do?

I'll create another PR to add validation formats to the BRProcessoField too and add tests to it, if that's ok.

Thanks.

claudep:
I'm investigating the issue about test failures. But it's unrelated to this patch, that's why I merged it.

LuanP:
Awesome! Thanks.

PR #174, opened 2015-10-18
added tests for BRProcessField and improved it with a regex

LuanP:
Also updated the link to where the validation can be found.


avelino:
very very good @LuanP

PR #175, opened 2015-10-24
Fix incorrect usage of cedilla instead of comma below for ș and ț letters in Romania county names

mnemonicflow:
Details:
* [Romanian alphabet](https://en.wikipedia.org/wiki/Romanian_alphabet)
* [Comma-below (ș and ț) versus cedilla (ş and ţ)](https://en.wikipedia.org/wiki/Romanian_alphabet#Comma-below_.28.C8.99_and_.C8.9B.29_versus_cedilla_.28.C5.9F_and_.C5.A3.29)


benkonrath:
@mnemonicflow Thanks for this. I can see you're pushing new versions (which is fine). Can you add a comment when it's ready so I can review and merge it? Thanks.

mnemonicflow:
@benkonrath Sorry for that. I initially forgot to update the tests. I think it's ready for review now.

benkonrath:
@mnemonicflow No problem at all. :) Thanks for this fix!

PR #177, opened 2015-10-27
Change regex of BRZipCodeField

gilsondev:
I'm change the regex to support many types of numbers:

 - 71000-000
 - 71.000-000
 - 71000000

### DIFF
--- a/localflavor/br/forms.py
+++ b/localflavor/br/forms.py
@@ -35,11 +35,11 @@ class BRZipCodeField(RegexField):
     XXXXX-XXX.
     """
     default_error_messages = {
-        'invalid': _('Enter a zip code in the format XXXXX-XXX.'),
+        'invalid': _('Enter a zip code in the format XXXXX-XXX, XX.XXX-XXX or XXXXXXXX.'),
     }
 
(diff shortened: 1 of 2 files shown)


claudep:
Please remove the .po/.mo changes from the patches. Those are handled separately through Transifex.

gilsondev:
Ok

gilsondev:
Removed

gilsondev:
Now, need access the django-localflavor in Transifex. I'm send a request to come in.

gilsondev:
@claudep files removed

claudep:
I'll let @avelino do the merge.

avelino:
Why validate point in CEP (zipcode of Brazil)?

gilsondev:
Yes, the point in zipcode is not default to use. I go remove this.

avelino:
@gilsondev I think this solution is bad, if a user puts a "," for example would break their implementation.

gilsondev:
So, you think better not change the pattern?

avelino:
Thinking about the use of lib, with this implementation not worth, if it is a generic implementation Yes.
Where did you get this need?

gilsondev:
I thought it would be interesting to have more than one format for zipcode

PR #180, opened 2015-11-12
Update test configs

benkonrath:


tox.ini
```diff
@@ -4,7 +4,7 @@ envlist =
     docs,
     py26-{1.5,1.6},
     {py27,py32,py33,py34}-{1.5,1.6,1.7,1.8},
-    {py27,py34}-master
+    {py27,py34,py35}-{1.9,master}
```
nagisa: This doesn’t match the TOXENV variables above.

benkonrath: Thanks. I'll fix it.

benkonrath:
It looks like python 3.5 is not available as `python3.5` in travis.

It seems like relying on all python binaries being available is not the best setup. For instance, it doesn't seem correct that the virturalenv is setup with python 2.7 in the python 3 tests.

python 3.5 test output.
https://travis-ci.org/django/django-localflavor/jobs/90712747
```
$ python --version
Python 2.7.9
$ pip --version
pip 6.0.7 from /home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages (python 2.7)
install.1
```

I'll investigate if there's a better way that this can be setup to get the python 3.5 testing.

benkonrath:
@nagisa This is ready to be reviewed. Thanks.

tests/settings.py
```diff
@@ -24,3 +24,5 @@
     TEST_RUNNER = 'discover_runner.DiscoverRunner'
 
 SECRET_KEY = 'spam-spam-spam-spam'
+
+MIDDLEWARE_CLASSES = ()
```
claudep: Any further details about this addition (apart from the commit message)?

benkonrath: Without this setting, the following system check warning is printed many times during the test run. Hiding this warning makes it easier to see the real errors and warnings (if there are any).

```
System check identified some issues:

WARNINGS:
?: (1_7.W001) MIDDLEWARE_CLASSES is not set.
	HINT: Django 1.7 changed the global defaults for the MIDDLEWARE_CLASSES.
    django.contrib.sessions.middleware.SessionMiddleware,
    django.contrib.auth.middleware.AuthenticationMiddleware, and
    django.contrib.messages.middleware.MessageMiddleware were removed from the  
    defaults. 
    If your project needs these middleware then you should configure this setting.
```

claudep: I see. Then I'd rather use something like https://github.com/carljm/django-model-utils/commit/f8a7c50c0ac4e4db9fa0e13f2834fcb364e361bb

nagisa:
This generally LGTM. Since @claudep seems to have some comments, I’ll defer reviewing/merging this to him.

benkonrath:
@claudep I didn't know about the `SILENCED_SYSTEM_CHECKS` option. I've updated the commit. Thanks.

claudep:
Looks good, thanks!

benkonrath:
Feel free to merge this. In general, I prefer other people to merge my PRs so there's a clear double check on the work. Thanks.

nagisa:
I’ll merge this since there’s clearly 2 r+ on this.

claudep:
The merge is fine, but when possible I generally prefer if the author merges (after review) so as the history is cleaner. I *hate* the unneeded merge commits that github makes.

nagisa:
@claudep doesn’t github always create the merge commit? AFAIK the only way to merge without the merge commit is to `--ff-only` merge the PR after fetching it locally.

claudep:
Yes, that was what I meant (rebase/push with fast forward).

benkonrath:
@claudep It would be nice to have some documentation of the preferred git workflow. I don't personally have a strong preference (besides what I mentioned earlier). I'm happy to adjust my workflow to match the project standard.

carljm:
@claudep I don't understand the link between merge-commit / fast-forward and who does the merge? Anyone with commit access can do either one.

I guess it's slightly easier for the author to do a rebase-ff because they already have the branch locally, is that what you mean?

claudep:
@carljm yes, that what I meant.
@benkonrath This is pure personal preference, not the project. You are an official maintainer now, so feel free to document anything if you feel the need.

benkonrath:
@claudep I was trying to say that maybe there should be a standard for the project for the sake of consistency. Since I don't have strong feelings about a particular workflow, I'm not the person to set this up. In any case, there are probably more important things to do ... like clear our PR backlog. :)

PR #186, opened 2015-11-25
Fix FRNationalIdentificationNumber validation for Corsica department

thomasdurin:
An error was made on the `FRNationalIdentificationNumber ` validation, department **20** (for Corsica) can't be ignored if birthdate is older than **1976**.

Before January 1976, department number for Corsica was 20. From 1976, Corsica has been divided into two number : 2A and 2B, as said on the [French departments page of wikipedia](https://fr.wikipedia.org/wiki/Département_français) : 
> In 1976, Corsica (number 20) was divided between the Corse-du-Sud (2A) and Haute-Corse (2B). (...) Social security numbers composition has changed for people born after 1 January 1976, the "20" has been replaced by "2A" or "2B".

Therefore we can't consider a French national identification number invalid if department number is **20** and birthdate **< 1976**.

For the record, this is not specified on the English version of the [French departments page of wikipedia](https://en.wikipedia.org/wiki/Departments_of_France).

### DIFF
--- a/tests/test_fr.py
+++ b/tests/test_fr.py
@@ -208,19 +208,38 @@ class FRLocalFlavorTests(SimpleTestCase):
         error_format = ['Enter a valid French National Identification number.']
         valid = {
             '869067543002289': '869067543002289',
-            '869069713002256': '869069713002256',   # Good Overseas
+            # Good Overseas
+            '869069713002256': '869069713002256',
+            # Good, old Corsica department number (20) with birthdate < 1976
+            '870062009002285': '870062009002285',
+            # Good, new Corsica department number (2A) with birthdate >= 1976
+            '882062A09002279': '882062A09002279',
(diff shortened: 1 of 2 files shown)


tests/test_fr.py
```diff
@@ -213,6 +213,8 @@ def test_FRNationalIdentificationNumber(self):
         invalid = {
             '369067543002289': error_format,    # Gender mismatch
             '869069873002289': error_format,    # Bad Department
+            '878062073002289': error_format,    # Bad Department with birthdate
+            '869062A73002289': error_format,    # Bad Department with birthdate
```
claudep: I would be more specific in the comment message, and also add some entries in the valid dictionary.

thomasdurin: Alright, I'll do this

thomasdurin:
@claudep Good for me, I added few tests and updated comments

PR #188, opened 2015-12-05
Adding some custom models fields in Brazil local flavor

andrelramos:
Include custom models fields for:
- Brazilian documents (CPF and CNPJ)
-  Telephone and Cell Phone
- Zip Code

I Also included testing for the same.


avelino:
There is a few steps before sending the PR.

- Open an issue documenting its proposal for change
- Don't break your tests
- Only commit for its amendment

tests/test_br.py
```diff
@@ -1,4 +1,3 @@
-# -*- encoding: utf-8 -*-
```
benkonrath: You need to re-add this line. I think this is the cause of the test failures.

benkonrath:
@avelino I'm not sure I understand your suggestions. Are you asking @andrelramos to open a separate issue to discuss the change before actually submitting the PR? Is there a specific reason you can't discuss this proposal on this PR? What do you mean by `Only commit for its amendment`? Thanks for helping me understand your reasoning so that I can follow along this review.

PR #189, opened 2015-12-09
Added some Polish translations.

m-strzelczyk:
Hello, I've added some missing translations for Polish language. I think now, most messages that aren't proper nouns are now translated.


claudep:
Hi Maciej, I'm sorry, but translations should be updated through Transifex, not by patch.
https://www.transifex.com/django/django-localflavor/
Then we will merge translations before publishing the next release.

PR #191, opened 2015-12-16
Add localflavor for Bulgaria

vstoykov:


benkonrath:
@vstoykov Thanks for your contribution. I should have time to review this tomorrow (Monday).

docs/changelog.rst
```diff
@@ -6,7 +6,7 @@ Changelog
 
 New flavors:
 
-- None
+- Added local flavor for Bulgaria (gh-191)
```
benkonrath: Can you add the github url like the other entries?

localflavor/bg/fields.py
```diff
@@ -0,0 +1,25 @@
+from django.db import models
+
+from .validators import egn_validator, eik_validator
+
+
+class EGNField(models.CharField):
```
benkonrath: All of the fields should be prefixed with the country code (in this case BG).

vstoykov: This means that the field's should be called `BGEGNField` and `BGEIKField` right?

localflavor/bg/validators.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def egn_validator(egn):
+    '''
+    Check Bulgarian unique citizenship number (EGN) for validity
+    More details https://en.wikipedia.org/wiki/Unique_citizenship_number
+    Full information in Bulgarian about algoritm is available here
+    http://www.grao.bg/esgraon.html#section2
+
+    '''
```
benkonrath: Can you use `"""` for this comment and remove the space below? This is for all comments in this PR.

localflavor/bg/validators.py
```diff
@@ -0,0 +1,68 @@
...
+    def check_checksum(egn):
+        _weights = (2, 4, 8, 5, 10, 9, 7, 3, 6)
```
benkonrath: I don't think the `_` is required for this variable name.

localflavor/bg/validators.py
```diff
@@ -0,0 +1,68 @@
...
+        try:
+            checksum = sum(weight * int(digit) for weight, digit in zip(_weights, egn))
+            return int(egn[-1]) == checksum % 11 % 10
```
benkonrath: Is it possible to express this a little more verbosely? I'm just thinking it would be easier to read if it didn't have the double modulus operator. This is just a suggestion though as there could be specific reason you're doing it like this.

localflavor/bg/validators.py
```diff
@@ -0,0 +1,68 @@
...
+        checksum = get_checksum(range(1, 9), eik)
+        if checksum == 10:
+            checksum = get_checksum(range(3, 11), eik)
+        return eik[8] == checksum % 10
+
+    def check_eik_extra(eik):
+        digits = eik[8:12]
+        checksum = get_checksum([2, 7, 3, 5], digits)
+        if checksum == 10:
+            checksum = get_checksum([4, 9, 5, 7], digits)
```
benkonrath: I think it's better to use a tuple instead of a list in these calls to `get_checksum`

localflavor/bg/validators.py
```diff
@@ -0,0 +1,68 @@
...
+            return None
+
+    if not (len(egn) == 10 and check_checksum(egn) and check_valid_date(egn)):
+        raise ValidationError(_("The EGN is not valid"))
+
+
+def eik_validator(eik):
+    """
+    Check Bulgarian EIK/BULSTAT codes for validity
+    full information in Bulgarian about algoritm is available here
```
benkonrath: There's a typo in `algoritm`. The same applies to another comment.

localflavor/bg/fields.py
benkonrath: That was my intention but it's hard to read with all the capital letters. Is there a way to expand `EGN` and `EIK` that will make sense?

benkonrath:
@vstoykov Thanks for adding nice comments and for adding the URLs - it's really helpful.

Besides the minor comments I've added, I noticed that not all code paths are tested. Is it possible to add specially crafted values to test the various checks you have?

localflavor/bg/fields.py
vstoykov: The naming of these fields was one of the hardest things. In Bulgaria:

-  ЕГН (transliterated EGN) stands for Единен Граждански Номер, which translated means Uniform Civil Number. 
- ЕИК (transliterated EIK) stands for Единен Идентификационен Код, which translated means Uniform Identity Code

The only way in English I saw ЕГН was in Wikipedia and it was just transliterated EGN. About ЕИК I never saw it written in English, and for that reason I just transliterated it also.

I can make the fields `BGUniformCivilNumber` and `BGUniformIdentityCode` if you prefer. I hope that someone will understand what are they because in the doc strings it is described.

vstoykov:
I tested only the validators because main logic is there, but for 100% coverage I can add test of the fields.

All changes that that I will do will be amended to this commit in order not to pollute the history with unnecessary commits.

localflavor/bg/fields.py
benkonrath: It looks like wikipedia calls it "uniform civil number" and  "Unique citizenship number". Would it make sense to use one of those and the English translation of `EIK`? Or do you think using the acronyms in the field names is better?

localflavor/bg/validators.py
vstoykov: Reason for doing double modulus operator are points 3. and 4. for calculating the checksum:

   1. Each digit is multiplied by its weight (see below)
   2. The products obtained are added
   3. The sum is divided by 11 ( use sum % 11, not /, modulus, not division)
   4. If the remainder is less than 10, the remainder is the checksum digit, otherwise the checksum digit is zero

When I first do modulus by 11 then result is not greater than 10 and instead of checking that I can just calculate modulus by 10.

localflavor/bg/fields.py
vstoykov: The reason I chose acronym is because if you say to some Bulgarian "Uniform civil number" or "Unique citizenship number" (in English) need to think a lot what this means. If you just say "EGN" everybody (in Bulgaria) will understand. I'm not sure if anyone outside of Bulgaria will use this field. If they really need it probably will know what EGN means.

 EGN in Bulgaria is similar to UMCN (Unique Master Citizen number) in countries that were created after the dissolution of Yugoslavia (Bosnia and Herzegovina, Macedonia, Montenegro, Serbia and Slovenia).
It is similar but the term UMCN is assigned to these countries and the algorithm in Bulgaria is different. For that reason I cant use this name also.

About EIK it is similar to EGN but only for legal personalities.

As I said in my earlier comment I can use the name `BGUniformCivilNumberField` instead of `EGNField` and `BGUniformIdentityCodeField` instead of `EIKField`.

vstoykov:
Ok I see now that not only fields are not checked but there are lines in validators which are not tested. I will look for test samples that will trigger respected lines to increase the coverage.

vstoykov:
@benkonrath I made changes for all your comments and updated my pull request.

localflavor/bg/fields.py
benkonrath: Sorry, I didn't see your response before I added the note about the wikipedia name which is why it seems like I didn't understand your response. 

The field name should primarily make sense to Bulgarian developers using it. It's also nice to have the `BG` prefex for consistency with other localflavors. The prefix also allows non-Bulgarian developers to quickly see that it's a field for Bulgaria. With these goals in mind, you are in the best position to choose between the English translation of the Bulgarian names or the acronyms (`BGEGNField`, `BGEIKField`). It seems to me that the acronyms are a better choice.

Thanks for the explanation.

localflavor/bg/validators.py
benkonrath: Ok, thanks for the explanation.

benkonrath:
@vstoykov Thanks for improving the tests and for making the other changes.

You need to add an app label the test models for your tests to work with Django >= 1.9
```
    class Meta:
        app_label = 'test_bg'
```
You can make a `test_bg` module if you want and include the test models in `models.py`.

I just noticed another small issue. The model fields should be in `models.py` instead of `fields.py` to match the other model fields in localflavor. 

Other than these 2 minor points, you just need to decide the best name for the fields. Thanks for your quick response on this.

vstoykov:
Ok I renamed `fields.py` to `models.py`, and both fields are now called `BGEGNField` and `BGEIKField`. Test models are fixed to work for Django>=1.9

benkonrath:
Looks good. One last thing: Can you add an entry with your name to `docs/authors.rst`? I'll merge this as soon as that's done. Thanks.

vstoykov:
Ok I updated my pull request with my name in `docs/authors.rst` :)

benkonrath:
@vstoykov Thanks for your work on this. I really appreciate your quick responses and that you took the time to answer my questions and generally explain things to me.

PR #201, opened 2016-02-15
Fixed PERUCField using RegexField instead of CharField

fmartingr:
In `localflavor.pe.forms.PERUCField` the `__init__` call had the wrong arguments. The first is a regular expression as in the other forms and the django API.

Using this form as it is raises an `AttributeError`: 

``` traceback
  File "/home/vagrant/.Envs/a/lib/python2.7/site-packages/localflavor/pe/forms.py", line 70, in clean
    value = super(PERUCField, self).clean(value)
  File "/home/vagrant/.Envs/a/lib/python2.7/site-packages/django/forms/fields.py", line 152, in clean
    self.run_validators(value)
  File "/home/vagrant/.Envs/a/lib/python2.7/site-packages/django/forms/fields.py", line 135, in run_validators
    v(value)
  File "/home/vagrant/.Envs/a/lib/python2.7/site-packages/django/core/validators.py", line 49, in __call__
    if not (self.inverse_match is not bool(self.regex.search(
AttributeError: 'int' object has no attribute 'search'
```

### DIFF
--- a/localflavor/pe/forms.py
+++ b/localflavor/pe/forms.py
@@ -7,7 +7,7 @@ from __future__ import unicode_literals
 
 from django.core.validators import EMPTY_VALUES
 from django.forms import ValidationError
-from django.forms.fields import RegexField, CharField, Select
+from django.forms.fields import CharField, Select
 from django.utils.translation import ugettext_lazy as _
 
 from .pe_region import REGION_CHOICES
@@ -49,7 +49,7 @@ class PEDNIField(CharField):
         return value
 
 
-class PERUCField(RegexField):
+class PERUCField(CharField):
     """
     This field validates a RUC (Registro Unico de Contribuyentes). A RUC is of
     the form XXXXXXXXXXX.


claudep:
If no regex is defined, then there are no reason the field inherits from `RegexField`. The correct fix might be to inherit from `CharField` instead.

fmartingr:
I think they used the `RegexField` thinking about giving the validator later, but after some little research I can't find information about the Peruvian ID number.

I'm changing the PR to `CharField` as proposed until we get more information about this.

fmartingr:
Searching it as RUC instead of its local name I found [this](https://github.com/robertozoia/validate-ruc/blob/master/validate_ruc.py) but I prefer if someone with knowledge in this can give an accurate answer.

claudep:
I think the patch is now correct. Proper validation of this field could be the subject of a later patch. @benkonrath, may I let you merge?

benkonrath:
@fmartingr Looks good. Can you rebase this into 1 commit?

It would be great if you could also add the validation but it's fine to do that in a different PR if you want.

The code you referenced is licensed under the MIT license which seems ok to use in this project.
http://www.dwheeler.com/essays/floss-license-slide.html

Just reference that repo if you end up using it for your implementation of the validation. That said, it would be better if you can find a more official source for the validation algorithm and make your own implementation. Spanish is ok for referencing the algorithm if you can't find documentation in English.

fmartingr:
I've rebased as requested.

If I end up doing the validation, I will make a new PR with that. People in my company should know how to validate the RUT code so I hope I get some official documents about that.

benkonrath:
Thanks.

PR #207, opened 2016-04-02
explicit requirements specifications

vladimirnani:


requirements.txt
```diff
@@ -0,0 +1 @@
+django
```
benkonrath: This should have a version. We currently support django>=1.7.

.travis.yml
```diff
@@ -40,7 +40,7 @@ cache:
     - $HOME/.wheels
 install:
     - mkdir -p $PIP_WHEEL_DIR
-    - pip wheel -r tests/requirements.txt codecov
+    - pip wheel -r requirements-tests.txt codecov
     # virtualenv>=14.0.0 has dropped Python 3.2 support
```
benkonrath: @vladimirnani Did you rename this to make it easier to find? Just curious.

requirements-tests.txt
```diff
@@ -2,3 +2,5 @@ invoke==0.11.1
 coverage==3.7.1
 flake8==2.2.5
 six==1.10.0
+
+-r requirements.txt
```
benkonrath: This is breaking the tests because django gets specified twice. I we should just not include this.

.travis.yml
vladimirnani: Moved to have dependency between requirements-tests and requirements.

requirements-tests.txt
vladimirnani: ok got it, so maybe we could have something like requirements-dev

benkonrath:
Fixed via bb1a1410c498f355ed0148a2f2177daa24cb7348. Thanks @vladimirnani!

PR #211, opened 2016-04-02
Better ZIP code validation for German ZIP codes.

daniponi:
Regex expression taken from:
http://www.fadoe.de/blog/archives/262-Regulaerer-Ausdruck-fuer-Deutsche-Postleitzahlen.html


Dunedan:
I can confirm that there is currently no German postal code starting with 00, so the adjusted regex should be fine, although I'm wondering if there is a formal definition that this postal code range will never be allocated. In any case it's quite unlikely that this happens and even if the regex can be simply adjusted again.

daniponi:
Gonna push this in a new branch

benkonrath:
See #216.

PR #215, opened 2016-04-02
Extra attributes problem, fixes #202

### LINKED ISSUE #202: IBANField with Nordea extensions does not validate
Using an IBANField with Nordea extensions enabled does not validate when trying to use a Nordea IBAN e.g. from Ukraine, UA573543470006762462054925026

Tested wtih:

Django 1.9.4
django-localflavor 1.2

**models.py**
```python
from django.db import models

from localflavor.generic.models import IBANField


class Partner(models.Model):
    iban = IBANField(use_nordea_extensions=True)
```

**admin.py**
```python
from django.contrib import admin

from .models import Partner


@admin.register(Partner)
class PartnerAdmin(admin.ModelAdmin):
    pass
```

ibanfield

vladimirnani:
Problem was that attributes were not passed to formfield validators, only to field validators.

### DIFF
--- a/tests/test_generic/models.py
+++ b/tests/test_generic/models.py
@@ -1,0 +1,11 @@
+from django.db import models
+
+from localflavor.generic.models import IBANField
+
+
+class UseNordeaExtensionsModel(models.Model):
+    iban = IBANField(use_nordea_extensions=True)
+
+
+class UseIncludedCountriesModel(models.Model):
+    iban = IBANField(include_countries=['NL'])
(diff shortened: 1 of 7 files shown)


tests/test_generic/models.py
```diff
@@ -0,0 +1,11 @@
+from django.db import models
+
+from localflavor.generic.models import IBANField
+
+
+class Partner(models.Model):
```
benkonrath: Can you rename this to something like `UseNordeaExtentions` to match `IncludedCountriesModel`? Same applies to the form.

benkonrath:
This looks good. Can you rename the classes that I indicated and then rebase / squash your commits into one? Thanks.

benkonrath:
@vladimirnani Thanks!

PR #218, opened 2016-04-02
Drop support for django 1.7.

benkonrath:
Django 1.7 isn't supported anymore so I don't think that we should continue to test against it. Thoughts / comments are welcome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,6 @@ matrix:
     - python: 3.5
       env: TOXENV=isort
     - python: 2.7
-      env: TOXENV=py27-1.7
-    - python: 2.7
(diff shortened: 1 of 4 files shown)


benkonrath:
After discussing this with some localflavor users at the DjangoCon EU sprints, we decided it was better to continue to support older versions of Django unless there's a specific reason we need to drop support for that version. For example, it would be ok to drop support for an older version of Django if there's a backwards incompatible change in Django that would require us to support separate code paths for the older version. I'm closing this PR.

benkonrath:
The deprecation work in #262 has problems with Djanog 1.7 so I've reopened this PR.

PR #219, opened 2016-04-03
Add a pull request review checklist to CONTRIBUTING.

benkonrath:
This is a checklist that I've developed for reviewing PRs. I'd like to publish so we can have a reference for contributors to look at before they submit a PR.

Let me know if there's a better place for this or if you have any suggestions. Thanks.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -59,5 +59,24 @@ And combine both options::
 
     $ COUNTRY=us tox -e py27-1.7
 
+Pull Request Review Checklist
+=============================
+
+- [ ] Prefix the country code to all fields.
+- [ ] Field names should be easily understood by developers from the
+      target localflavor country. This means that English translations
+      are usually not the best name unless it's for something standard
+      like postal code, phone number, tax / VAT ID etc.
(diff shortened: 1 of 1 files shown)


nagisa:
Similarly to my suggestion to use bots for finding out maintainers, I think many of these checkpoints can (and should) be automatically checked for by some sort of automated means (travis?).

benkonrath:
I agree that automating this kind of thing is good. It's sometimes hard to justify spending time on the automation for a small project like localflavor but it's something to work towards.

In the meantime, I'm going to add this to a pull request template instead of adding this to CONTRIBUTING.
https://github.com/blog/2111-issue-and-pull-request-templates

benkonrath:
Fixed via https://github.com/django/django-localflavor/commit/cad42c2eed9e69908f12124f4e246b79b8b10f57.

benkonrath:
@nagisa Do you know of tool we can use to check for the `..version_added` markers?

nagisa:
@benkonrath the most trivial way to do it would be to have a script which checks for existence of it on newly added API and run the script during CI.

PR #223, opened 2016-04-09
WIP: Deprecate all phonenumber validators and model fields.

orf:
This is a WIP MR that deprecates all phone-number validators, issuing a warning suggesting people switch to `django-phonenumber-field`.

Why? It seems pointless to keep maintaining the phone number validators when they are incomplete and/or inferior to the comprehensive list of validators `django-phonenumber-field` (and `libphonenumber` behind it) provide. A good way forward is to suggest that people migrate to that package, or alternatively implement our own phone number validators based on `libphonenumber`.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -4,6 +4,12 @@ Changelog
 1.3   (unreleased)
 ------------------
 
+Deprecations:
+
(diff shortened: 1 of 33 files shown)


benkonrath:
Closed in favour of #262.

PR #224, opened 2016-04-10
Add deconstruct to model fields

benkonrath:
This builds on the @mlissner's commit in PR #162. I've added a deconstruct method to all model fields. I've also added a generic test for the deconstruct as well as a test for the options in the IBAN Field. It would be great if somebody could review this and provide feedback. Thanks.

Closes #162

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -50,6 +50,7 @@ Authors
 * Marti Raudsepp
 * Martin Ogden
 * Matias Dinota
+* Mike Lissner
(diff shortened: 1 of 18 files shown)


benkonrath:
The Python 3.2 test failure only happens with the Travis version of Python 3.2.5. I can't reproduce the problem on my local system with Python 3.2.6. I need to think about how I can get around this but the general PR is still ready for feedback.

benkonrath:
Tests are now passing.

PR #225, opened 2016-04-11
Change the french regions

dduong42:
Some regions have been merged

See http://www.insee.fr/fr/methodes/nomenclatures/cog/telechargement/2016/txt/reg2016.txt and https://fr.wikipedia.org/wiki/Loi_relative_%C3%A0_la_d%C3%A9limitation_des_r%C3%A9gions,_aux_%C3%A9lections_r%C3%A9gionales_et_d%C3%A9partementales_et_modifiant_le_calendrier_%C3%A9lectoral


claudep:
Thanks for the patch. Even if the French localflavor doesn't provide a model based on those regions, we can imagine that some user of the library may. That user should then migrate the values in her database. I think we should at least add a note about this change in the docs, and ideally provide an example of a migration function.

benkonrath:
@dduong42 Are you still interested in contributing this change? If so, please consider implementing claudep's suggestions. Thanks.

benkonrath:
I'm closing PR because of inactivity. The changes here may have been integrated in #260. If the changes are still required, feel free to re-open this PR.

PR #231, opened 2016-05-07
Adding support for Kuwait governorates

burhan:
This PR adds a select field that enumerates the 6 governorates of Kuwait; as per [ISO 3166-2](https://en.wikipedia.org/wiki/ISO_3166-2:KW)

Note that in the wikipedia entry they are listed as _subdivision_ but they are actually governorates.

### DIFF
--- a/localflavor/kw/kw_governorates.py
+++ b/localflavor/kw/kw_governorates.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.utils.translation import ugettext_lazy as _
+
+# List of governorates as defined by ISO_3166-2
+# For reference; see https://en.wikipedia.org/wiki/ISO_3166-2:KW
+# Note that these are governorates and not subdivisons
+
+# These are the English names
+GOVERNORATE_CHOICES = (
+    ('AH', _('Ahmadi')),
+    ('FA', _('Farwaniyah')),
+    ('JA', _('Jahra')),
+    ('KU', _('Capital')),
+    ('HA', _('Hawalli')),
+    ('MU', _('Mubarak Al Kabir')),
+)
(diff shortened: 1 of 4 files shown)


localflavor/kw/kw_governorates.py
```diff
@@ -0,0 +1,18 @@
...
+# Note that these are governorates and not subdivisons
+
+# These are the English names
+GOVERNORATE_CHOICES = (
+    ('AH', _('Ahmadi')),
+    ('FA', _('Farwaniyah')),
+    ('JA', _('Jahra')),
+    ('KU', _('Capital')),
+    ('HA', _('Hawalli')),
+    ('MU', _('Mubarak Al Kabir')),
```
benkonrath: I think it would be better to use the full ISO_3166-2:KW code for these (e.g.' KW-AH').

localflavor/kw/kw_governorates.py
```diff
@@ -0,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.utils.translation import ugettext_lazy as _
+
+# List of governorates as defined by ISO_3166-2
+# For reference; see https://en.wikipedia.org/wiki/ISO_3166-2:KW
+# Note that these are governorates and not subdivisons
```
benkonrath: There's a typo in the comment: 'subdivisons' -> 'subdivisions'.

Does this comment just meant that the name for these boundaries is 'governorate', not 'subdivision'? Or are governorates something different than subdivisions?

localflavor/kw/kw_governorates.py
burhan: I wouldn't mind that - except in Real Life - the KW- is dropped as its redundant.

localflavor/kw/kw_governorates.py
burhan: Governorates are different than subdivisions, in that subdivisions don't have governors, and governorates do.

So, in Kuwait there is a Governor of Ahmadi

The spelling mistake, I'll fix that ASAP.

localflavor/kw/kw_governorates.py
benkonrath: Sounds good. Thanks.

PR #232, opened 2016-05-07
refactoring field to inherit from RegexField

burhan:
As discussed in #195 - this PR refactors the field to inherit from RegexField; further it simplifies the regular expression as well.

### DIFF
--- a/localflavor/kw/forms.py
+++ b/localflavor/kw/forms.py
@@ -3,25 +3,39 @@ Kuwait-specific Form helpers
 """
 from __future__ import unicode_literals
 
+import textwrap
 import re
 from datetime import date
 
 from django.core.validators import EMPTY_VALUES
 from django.forms import ValidationError
-from django.forms.fields import Field, Select
+from django.forms.fields import RegexField, Select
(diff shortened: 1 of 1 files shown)


localflavor/kw/forms.py
```diff
@@ -3,23 +3,29 @@
...
 from django.utils.translation import gettext_lazy as _
 
-id_re = re.compile(r'''^(?P<initial>\d{1})
-                       (?P<yy>\d\d)
-                       (?P<mm>\d\d)
-                       (?P<dd>\d\d)
-                       (?P<mid>\d{4})
-                       (?P<checksum>\d{1})''', re.VERBOSE)
 
+def has_valid_checksum(value):
```
benkonrath: Since you're making this a public function, you should rename this to be more meaningful.  `kw_civil_id_number_has_valid_checksum` could work but I'm also open to other suggestions.

benkonrath:
@burhan Can you rebase this with master so that it can be merged? Thanks.

burhan:
rebased

PR #235, opened 2016-05-18
Change DIF to CDMX in mx_states

svmalvaez:
Thanks for your contribution!

first version. You'll be asked to rebase and squash the commits before the
final merge.

Please replace these instructions with a description of your change. The
'New Fields Only' checkboxes should be removed as well if your pull request
doesn't add any new fields.

**All Changes**
- [ ] Add an entry for your name in the docs/authors.rst file if it's not
    already there.
  
**New Fields Only**
    tax / VAT ID etc.
    States postal system.
    localflavors.


benkonrath:
@svmalvaez Please edit the description of this PR to add a description of your change along with the relevant check box items.

localflavor/mx/mx_states.py
```diff
@@ -9,11 +9,11 @@
     ('BCN', _('Baja California')),
     ('BCS', _('Baja California Sur')),
     ('CAM', _('Campeche')),
+    ('CDMX', _('Ciudad de México')),
```
benkonrath: The state field is 3 characters long this won't work. You'll also need to modify the tests for this change.

jackboot7: Would be using the three letter airport code for Ciudad de México (MEX) a fix for this?

sebasmagri: @jackboot7 the official abbreviation is CDMX, so it would probably need the state field to be updated.

elyak123: Also I believe it can be confusing with "Estado de México" which is a different state, I know its not very conventional but as a compromise could it be `CMX` ??

PR #236, opened 2016-05-25
Allow passing field name as first positional argument

emesik:
The convention for Django model fields is to pass field's verbose name as the first positional argument. However, the signature of localflavor.generic.models.IBANField.__init__ is incompatible with that.

This fix makes the field follow the convention.


benkonrath:
Thanks for your contribution. I fixed the failing test and committed this change via 15b4b9576eb4355ac5e03d1c639fbe26eed28b5d.

PR #239, opened 2016-07-11
Add birth date check to PESEL validation

emesik:
The PESEL number contains birth date encoded within first 6 digits. Checksum validation is enough to spot most typos and swapped digits but doesn't protect from injecting numbers with invalid month or day value.

The new validator decodes the birth date components and checks by calling datetime.date() on them.

### DIFF
--- a/localflavor/pl/forms.py
+++ b/localflavor/pl/forms.py
@@ -4,6 +4,7 @@ Polish-specific form helpers
 
 from __future__ import unicode_literals
 
+import datetime
 import re
 
 from django.core.validators import EMPTY_VALUES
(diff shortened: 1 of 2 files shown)


localflavor/pl/forms.py
```diff
@@ -38,12 +39,14 @@ class PLPESELField(RegexField):
...
         * the length consist of 11 digits
         * has a valid checksum
+        * contains a valid birth date
 
     The algorithm is documented at http://en.wikipedia.org/wiki/PESEL.
     """
     default_error_messages = {
         'invalid': _('National Identification Number consists of 11 digits.'),
         'checksum': _('Wrong checksum for the National Identification Number.'),
+        'birthdate': _('The number contains invalid birth date.'),
```
benkonrath: Can you make this message match the other messages? Here's my suggestion:
```
 The National Identification Number contains an invalid birth date.
```

PR #240, opened 2016-07-26
BRCNPJField is now a CharField

poiati:
Make the `localflavor.br.forms.BRCNPJField` a subclass of `django.forms.CharField` (instead of `django.forms.Field`). 

With this change the user can select with formats the field can accept. E.g:

Both formats bellow are valid:
- 56.352.556/0001-00
- 56352556000100

But with the current implementation the user can't select just one of them. With this he can using the `min_length` and `max_length` from `django.forms.CharField`.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -62,6 +62,7 @@ Authors
 * Michał Sałaban
 * Mike Lissner
 * Olivier Sels
+* Paulo Poiati
 * Rael Max
 * Ramiro Morales
 * Rolf Erik Lekang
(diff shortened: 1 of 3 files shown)


benkonrath:
@poiati Thanks for your contribution and sorry for the delay in getting to this - I was on vacation. The PR looks good, it just needs a test for the newly supported format. Thanks.

poiati:
Hello @benkonrath, it isn't a new format. The field always accepted two formats:

- With punctuations;
- Without punctuations.

This PR makes possible allowing only one of those formats, and it do this through the usage of the `max_length` and `min_length` from Django `CharField` (that already has tests).

benkonrath:
@poiati I understand now. Thanks for the clarification. Can you add a sentence to the docstring about this just so that people know that this option is available? Thanks.

benkonrath:
Fixed via #264.

PR #243, opened 2016-09-28
Adding local flavor for Venezuela

jackboot7:
Added local flavor for Venezuela data. 

- Added regions choice data.
- Added Select widget for regions.


localflavor/ve/ve_region.py
```diff
@@ -0,0 +1,23 @@
...
+"""
+A list of Venezuelan regions for use as `choices` in a formfield.
+
+Data based from http://en.wikipedia.org/wiki/Regions_of_Venezuela
+"""
+
+#: A list of Venezuelan regions for use as `choices` in a formfield.
+REGION_CHOICES = (
+    ('501', _(u'Región Capital')),
+    ('502', _(u'Región Central')),
```
claudep: Please remove the u'' prefixes. They are not necessary as you import unicode_literals anyway.

jackboot7: Done. Updated PR.

claudep:
Thanks for the contribution, but the patch will also need tests.

jackboot7:
Will fix the errors and send a new one.

PR #246, opened 2016-10-05
Added MXPhoneNumberField model and form fields.

jrperdomoz:
**All Changes**
    already there.
  
**New Fields Only**
    tax / VAT ID etc.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -46,6 +46,7 @@ Authors
 * Jocelyn Delalande
 * Johnny Lee Othon
 * Jonas Ghyllebert
+* Jonatan Rodríguez
 * Joseph Kocherhans
 * Josh Crompton
 * Julien Phalip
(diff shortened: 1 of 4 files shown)


benkonrath:
Thanks for your contribution. Unfortunately we're in the process of depreciating all phone number fields in localflavor in favour of django-phonenumber-field.

https://github.com/stefanfoulis/django-phonenumber-field

django-phonenumber-field is actively updated for phone number changes and has global support where as our phone number fields are lagging behind phone number changes and have bugs. I feel a bit bad about this because I know you spent time on this but it's unlikely that this PR will be merged. 

Are you able to use django-phonenumber-field for your needs? If so, please close the PR. 

Thanks again for your contribution. I really appreciate that you took the time to read the instructions and make a proper PR using our guidelines. I'm looking forward to seeing your next contribution to localflavor.

jrperdomoz:
Don't worry.  Good Job!

PR #250, opened 2016-11-05
Stop the translation for state 'georgia' and the country from colliding

### LINKED ISSUE #249: US: Translation of the state Georgia collides with the country
Turns out there's an US state Georgia and a country with both the exact same name, but, at least in Germany, only the country Georgia has a translation. So translating the US state to the same thing as the country causes confusion.
I have "fixed" this in our project using

``` python
US_STATES_GEORGIAFIX = map(lambda state: ('GA', "Georgia") if state[0] == 'GA' else state, US_STATES)
```

More generally asking: Should US state names be translated at all?

orf:
I removed the seemingly redundant duplication of US states into a list of contiguous and non-contiguous states, then simply combine them. I sorted them by the state name to keep the ordering we had before.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -76,5 +76,6 @@ Authors
 * Tino de Bruijn
 * Trey Hunner
 * Tyler Ball
+* Tom Forbes
 * Venelin Stoykov
 * Vladimir Nani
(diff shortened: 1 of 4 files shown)


benkonrath:
Thanks @orf! The next step is to figure get the translations updated so that this can be fixed for German.

claudep:
cd localflavor; django-admin makemessages -l en
than commit, push, and Transifex should pick the new file after some hours.

benkonrath:
@claudep  Thanks. I didn't realize that Transifex automatically picked up the translations and I could do this myself.

PR #252, opened 2016-11-05
Finish refactoring field to inherit from RegexField 

benkonrath:
Picking up the PR in #232.

Closes #232.


benkonrath:
This has been fixed in #232. No need for my fix. Thanks @burhan!

PR #253, opened 2016-11-05
Use Travis cache to speed up test runs.

orf:
Use `cache: pip` to cache between runs


benkonrath:
For future reference, here's the documentation for this feature:

https://docs.travis-ci.com/user/caching/#pip-cache

PR #260, opened 2016-11-06
Added new french regions (2016 reform).

alix-:
Added new french regions dataset (http://www.insee.fr/fr/methodes/nomenclatures/cog/).

We should keep the old regions and refer to these ones as "new regions", as the government's decision to change french regions is quite controversial.
Therefore, it's a good idea to leave the option to django-localflavour users to keep using the old regions system, at least for a little while.

**All Changes**

      already there.

**New Fields Only**

      tax / VAT ID etc.

### DIFF
--- a/tests/test_fr.py
+++ b/tests/test_fr.py
@@ -4,8 +4,8 @@ from __future__ import unicode_literals
 from django.test import SimpleTestCase
 
 from localflavor.fr.forms import (FRDepartmentField, FRDepartmentSelect, FRNationalIdentificationNumber,
-                                  FRPhoneNumberField, FRRegionField, FRRegionSelect, FRSIRENField, FRSIRETField,
-                                  FRZipCodeField)
+                                  FRNewRegionSelect, FRPhoneNumberField, FRRegionField, FRRegionSelect, FRSIRENField,
+                                  FRSIRETField, FRZipCodeField)
 
 
 DEP_SELECT_OUTPUT = '''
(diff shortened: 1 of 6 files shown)


benkonrath:
@claudep I feel you're more qualified than I am to review this PR. Do you have time for this review? Thanks.

claudep:
As a real French citizen, @aaugustin might be a better reviewer :-)

tests/test_fr.py
```diff
@@ -156,6 +156,29 @@
...
+        <option value="03">03 - Guyane</option>
+        <option value="04">04 - La Réunion</option>
+        <option value="06">06 - Mayotte</option>
+        <option value="11">11 - Île-de-France</option>
+        <option value="24">24 - Centre-Val de Loire</option>
+        <option value="27">27 - Bourgogne-Franche-Comté</option>
+        <option value="28">28 - Normandie</option>
+        <option value="32">32 - Nord-Pas-de-Calais-Picardie</option>
+        <option value="44">44 - Alsace-Champagne-Ardenne-Lorraine</option>
+        <option value="52" selected="selected">52 - Pays de la Loire</option>
```
aaugustin: Why is this a particular region preselected by default? This one isn't a bad choice, but wouldn't it be best not to select any?

aaugustin:
Indeed France messed up with regions once again and this patch correctly reflects the latest version of the mess. The data is all right.

@claudep This looks ready for checkin to me. 

@alix- Thanks!

benkonrath:
@alix- Thanks for the PR. We're trying to keep only current region names in localflavor rather than keeping historical data as well. Here's some information the policy:
https://django-localflavor.readthedocs.io/en/latest/#backwards-compatibility
We'll be switching to the django warnings system shortly so that we can deprecate fields and provide help with migrating data (there are PRs open for this).

For this case, do think there's a use case for keeping the existing data at least for a little while? If so, then this PR seems like a good option. At some point we could depreciate the original field and remove it. I'm curious to hear people's thoughts on this. Thanks.

alix-:
@benkonrath Yes, I think there is value in keeping the "old" regions for a while.

- The French are not yet used to these regions, even though these are the official new regions.
- There is a LOT of political controversy regarding these changes and I can see developers from certain parts of France not wanting to use this new system as a sign of protest.

aaugustin:
When I read the patch I wrote "yes keeping the old definition for backwards compatibility is smart". Then I noticed it was against the guidelines so I removed that comment. FYI ;-)

benkonrath:
Ok, great. Thanks again for your contribution!

batisteo:
:point_up: Wait, the names changed by June

[fr.wikipedia.org/wiki/Région_française#Liste_des_régions_actuelles](https://fr.wikipedia.org/wiki/R%C3%A9gion_fran%C3%A7aise#Liste_des_r.C3.A9gions_actuelles)

benkonrath:
@batisteo I don't understand. Is there something wrong with the merged code?

batisteo:
One example: I see **Languedoc-Roussillon-Midi-Pyrénées** instead of **Occitanie**.
The new regions are one year old, but the names wasn't fixed until this summer.

And it seems like the INSEE just "refactored" their site so the URL is not correct anymore.

benkonrath:
The best thing to do is make a new pull request against master so we can discuss your proposed update. Thanks.

batisteo:
See my raw pull request #268

PR #263, opened 2016-11-11
Refactored complex methods

### LINKED ISSUE #261: refactor complex methods
These methods are too complex to pass prospector so should be refactored

**mccabe: MC0001 / IDLicensePlateField.clean is too complex (12)**
localflavor/id_/forms.py
Line: 111

**mccabe: MC0001 / NOSocialSecurityNumber.clean is too complex (13)**
localflavor/no/forms.py
Line: 46

vladimirnani:
Specified complexity level for prospector.

Refactored methods that raised complexity exceptions. (max-complexity 9)

Fixes #261

### DIFF
--- a/localflavor/bg/validators.py
+++ b/localflavor/bg/validators.py
@@ -1,10 +1,12 @@
 from django.core.exceptions import ValidationError
+from django.utils.deconstruct import deconstructible
 from django.utils.translation import ugettext_lazy as _
 
 from .utils import get_egn_birth_date
 
 
-def egn_validator(egn):
+@deconstructible
+class EGNValidator(object):
     """
     Check Bulgarian unique citizenship number (EGN) for validity.
 
@@ -12,7 +14,8 @@ def egn_validator(egn):
     Full information in Bulgarian about algorithm is available here
     http://www.grao.bg/esgraon.html#section2
     """
-    def check_checksum(egn):
+
+    def _check_checksum(self, egn):
         weights = (2, 4, 8, 5, 10, 9, 7, 3, 6)
         try:
             checksum = sum(weight * int(digit) for weight, digit in zip(weights, egn))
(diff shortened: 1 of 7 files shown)


benkonrath:
Thanks! I should have time to review this tomorrow.

localflavor/bg/validators.py
```diff
@@ -39,30 +44,35 @@ def eik_validator(eik):
...
-    except ValueError:
-        raise ValidationError(error_message)
-
-    if not (len(eik) in [9, 13] and check_eik_base(eik)):
-        raise ValidationError(error_message)
 
-    if len(eik) == 13 and not check_eik_extra(eik):
-        raise ValidationError(error_message)
+eik_validator = EIKValidator()
+egn_validator = EGNValidator()
```
benkonrath: What do you think about deprecating the *_validator versions? I suppose there's no real harm in keeping them around but it would be nice to provide just the class version. Users can always make their own version of the callable if they want to use it that way. Consider this optional and only do it if you agree.

localflavor/bg/validators.py
```diff
@@ -1,36 +1,41 @@
...
     """
     Check Bulgarian unique citizenship number (EGN) for validity.
 
     More details https://en.wikipedia.org/wiki/Unique_citizenship_number
     Full information in Bulgarian about algorithm is available here
     http://www.grao.bg/esgraon.html#section2
     """
-    def check_checksum(egn):
+
+    def check_checksum(self, egn):
```
benkonrath: Can you mark new methods as private (prefixed with _)? This applies to all of the methods that you've created on the forms. I want to be clear that these are internal and not public. Please to this for data fields on the form as well.

localflavor/bg/validators.py
vladimirnani: Yes I was thinking to deprecate them. There is no point in callable, because usually logic is complex so you have to split into several methods. And even for simple cases think is better to stick to class validators conventionally.

benkonrath: I'm ok with merging this before these are depreciated because the deprecation work in #262 hasn't been merged yet. Just file a new issue so we don't forget to do the deprecation once #262 has been merged.

vladimirnani:
Thanks for review, have created [265](https://github.com/django/django-localflavor/issues/265).

benkonrath:
Thanks @vladimirnani!

PR #268, opened 2016-11-26
Renamed the new French regions (2016 reform).

batisteo:
The names of the new regions were adopted. Four names changed:
Hauts-de-France, Nouvelle-Aquitaine, Grand-Est and Occitanie.

Wikipedia source: [fr.wikipedia.org/wiki/Région_française#Choix_du_nom_des_nouvelles_régions_en_2016](https://fr.wikipedia.org/wiki/Région_française#Choix_du_nom_des_nouvelles_r.C3.A9gions_en_2016)

**All Changes**

      already there.

**New Fields Only**

      tax / VAT ID etc.

      States postal system.

      localflavors.


batisteo:
BTW, I kind of dislike using _new_. I would prefer using `REGION_2016_CHOICES` for example.
In few years it won't be new. And it might be new new regions in a while.

benkonrath:
@aaugustin @alix- Can you look over these proposed changes and give your opinion of using "2016" instead of the "new" in the names? Thanks.

alix-:
@batisteo Do you have any offical source for these changes ? Are they in effect today ?

According to Insee, the official regions are still the ones I implemented in #260 (https://www.insee.fr/fr/information/2114819#titre-bloc-30)

batisteo:
@alix- I've got this at the moment:
`curl 'https://www.insee.fr/fr/metadonnees/geographie-administrative/cog/suggestion' -d '{"q":"7","filters":[{"field":"geographie_codeNiveau","values":["REG"]}],"sortFields":[]}' -H 'Content-Type: application/json'`

batisteo:
And this: https://www.insee.fr/en/recherche/recherche-geographique?geo=REG-32+REG-44+REG-75+REG-76

alix-:
Looks good to me, the changes seem legit (always takes a while for these kind of changes to propagate across official websites).
I'm also in favour of changing _new_ to _2016_.

batisteo:
Maybe this as well: http://www.interieur.gouv.fr/Actualites/L-actu-du-Ministere/Les-noms-des-nouvelles-regions-sont-actes

alix-:
If you want these changes accepted, you should fix the failing tests (they check for a specific HTML output containing the regions names).

benkonrath:
I'm hoping to do the 1.4 release in a couple of weeks and it would be nice to get the name changes and the new -> 2016 change in before the release. Thanks.

benkonrath:
Thanks!

batisteo:
Thanks @benkonrath and @alix- for your quick feedback!

aaugustin:
A bit late to the party — this looks good. Thanks @batisteo.

PR #270, opened 2016-12-17
Added the Moroccan localflavor

daker:
**Changes**

* Added regions and provinces data.
* Added forms helpers for select widgets.
* ~~Added PhoneNumber field.~~
* Added PostalCode field.

**All Changes**

      already there.

**New Fields Only**

      tax / VAT ID etc.

- [x] Prefer '<country code>PostalCodeField' for postal codes as it's
      States postal system.

      localflavors.


moughamir:
Nice job!!

benkonrath:
Thanks for you contribution. We're in the process of deprecating the phone number fields in localflavor. We now recommend that people use https://github.com/stefanfoulis/django-phonenumber-field. Could you remove the phone number field after which I'll do a full review. Thanks!

claudep:
Generally, geographic names are marked for translation. I can suppose for example that one may want to translate the names to Arabic. It's not a requirement, just a suggestion.

benkonrath:
Thanks @claudep . This is my mistake, I'll correct it.

PR #273, opened 2016-12-27
Add local flavor for Ukraine.

illia-v:


localflavor/ua/ua_regions.py
```diff
@@ -0,0 +1,33 @@
+from django.utils.translation import ugettext_lazy as _
+
+#: 24 oblasts, Autonomous Republic of Crimea and 2 cities with special status
+# Codes were gotten from KOATUU
+UA_REGION_CHOICES = (
+    ('71', _('Cherkasy Oblast')),
```
benkonrath: I would prefer that you use [ISO_3166-2:UA](https://en.wikipedia.org/wiki/ISO_3166-2:UA) for the stored code instead of KOATUU. According to the [KOATUU English wikipedia page](https://en.wikipedia.org/wiki/Classification_of_objects_of_the_administrative-territorial_system_of_Ukraine), KOATUU will be integrated into ISO_3166-2:UA at some point in the future. Using the ISO standard also simplifies the discussion around disputed territories because we can just follow the standard. Does using ISO_3166-2:UA make sense to you? 

I should point out that not everybody will be happy with ISO_3166-2:UA in regards to Crimea and Sevastopol being part of Ukraine. The solution you've implemented should allow developers to monkey-patch a region list that is suitable for their target audience so I'm ok with this PR being merged with Crimea and Sevastopol in the list.

illia-v:
@benkonrath, thanks. I have changed what you requested.

benkonrath:
Thanks for switching to the ISO 3166-2:UA codes. I meant that only the codes should be changed, not the name of the regions. It's better to use the English version of the region names instead of the transliterated region names because English is the base language for translations. You can find the English names on this page:
https://en.wikipedia.org/wiki/Administrative_divisions_of_Ukraine#Regions

Sorry for the confusion. Does this make sense to you?

illia-v:
@benkonrath, changed that. Sorry for delay.

benkonrath:

PR #277, opened 2017-02-14
Add normalized versions of COFA state names

elliottcf:
I added the `COFA_STATES` names in a few forms to the `STATES_NORMALIZED` dict.   

We needed this so we could use the `localflavor.us.forms.USStateField` in a form which accepts COFA locations.  

I ran tox even though I was just adding a few more entries to an existing dict.


elliottcf:
I just squashed my commits to make it easier to merge. Let me know if there is anything else I should do.

benkonrath:
Thanks for your contribution. Could you update the comment in the `USStateField` to mention that the US territories and the COPA countries are also included in the validation.

For anybody who's following and interested, this video explains how the US territories and the COPA countries are organized.
https://www.youtube.com/watch?v=ASSOQDQvVLU

elliottcf:
I updated the `USStateField` comment to mention the additional territories it validates.  
I also squashed my commits into one.

benkonrath:

PR #278, opened 2017-02-20
Add Australian Company Number validator

koterpillar:
Add [Australian Company Number](https://en.wikipedia.org/wiki/Australian_Company_Number) model and form fields.

Note that running `isort --recursive --line-width 120 localflavor tests` produced changes in otherwise unchanged files. I've limited my changes to the files I touched.

**All Changes**

      already there.

**New Fields Only**

      tax / VAT ID etc.

- [x] Prefer '<country code>PostalCodeField' for postal codes as it's
      States postal system.

      localflavors.

### DIFF
--- a/tests/test_au/tests.py
+++ b/tests/test_au/tests.py
@@ -133,6 +133,18 @@ class AULocalflavorTests(TestCase):
         }
         self.assertFieldOutput(forms.AUBusinessNumberField, valid, invalid)
 
+    def test_acn(self):
+        error_format = ['Enter a valid ACN.']
+        valid = {
+            '604327504': '604327504',
+            '604 327 504': '604327504',
+        }
+        invalid = {
+            '604327505': error_format,
+            '60A327504': error_format,
+        }
+        self.assertFieldOutput(forms.AUCompanyNumberField, valid, invalid)
+
     def test_tfn(self):
         error_format = ['Enter a valid TFN.']
         valid = {
--- a/tests/test_au/tests.py
+++ b/tests/test_au/tests.py
@@ -322,6 +322,22 @@ class AULocalFlavourAUBusinessNumberFormFieldTests(TestCase):
         self.assertEqual('53 004 085 616', field.prepare_value('53 0 04 08561 6'))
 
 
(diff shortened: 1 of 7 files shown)


tests/test_au/tests.py
```diff
@@ -195,6 +196,56 @@ def test_raises_error_for_invalid_abn(self):
...
+
+    def test_raises_error_for_too_long_acn(self):
+        """Test an ACN with more than nine digits is invalid."""
+        invalid_acn = '6043275040'
+        validator = AUCompanyNumberFieldValidator()
+        self.assertRaises(ValidationError, lambda: validator(invalid_acn))
+
+    def test_raises_error_for_whitespace(self):
+        """Test an ACN can be valid when it contains whitespace."""
+        # NB: Form field should strip the whitespace before regex valdation is run.
```
benkonrath: valdation -> validation

I see that's also in the copy & pasted code so maybe you can also fix the typo there as well.

As this comment points out, the whitespace is stripped in the form. It would be nice to have a test for the form as well - at least testing the expected functionality that isn't tested with the validation tests.

tests/test_au/tests.py
```diff
@@ -195,6 +196,56 @@ def test_raises_error_for_invalid_abn(self):
...
 
+class AULocalFlavorAUCompanyNumberFieldValidatorTests(TestCase):
+
+    def test_no_error_for_a_valid_acn(self):
+        """Test a valid ACN does not cause an error."""
+        valid_acn = '604327504'
+
+        validator = AUCompanyNumberFieldValidator()
+
+        validator(valid_acn)
```
benkonrath: It would be nice to be consistent on the spacing. For instance, some tests have a space between each statement while one has no spaces. The space don't seem to be needed because it's pretty simple code. Feel free to fix the previous tests as well where this was copied from.

koterpillar:
Fixed both issues.

benkonrath:
It would be great if you can also add a test the form's `prepare_value`. I think there's an example test with the `AUBusinessNumberField`. Thanks.

koterpillar:
Added tests.

PR #280, opened 2017-03-06
Fix saving a single space for empty postcode value.

vdboor:
When the `NLZipCodeField` contains an empty string, it saves an empty space instead.
This PR updates the test code to check this, and provides the fix.

I've also applied `len(value)` logic found in the formfield `NLZipCodeField.clean()` so the models' `to_python()` becomes more rubust.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -27,6 +27,7 @@ Authors
 * Daniela Ponader
 * Danielle Madeley
 * Daniel Roschka
+* Diederik van der Boor
 * d.merc
 * Douglas Miranda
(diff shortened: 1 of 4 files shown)


jieter:
Looks good to me.

vdboor:
@benkonrath I've added the entry! :-)

I was a bit confused about the Chicken & Egg issue to add a changelog for a PR that I didn't submit yet.

benkonrath:
Yeah, you just have to follow up with a second commit for the changelog entry. Thanks again.

PR #295, opened 2017-06-14
Fixes #294

### LINKED ISSUE #294: Django 1.11.x and null=True Blank=Ture with us.models/forms.USZipCodeField causes crash
After upgrading from Django 1.10.x to Django 1.11.2 both using localflavor 1.5, the USZipCodeField with null=True, blank=True causes a crash during validation of a `ModelForm` when the zip code is left blank.

Model:
```
class EmergencyContact(models.Model):
   # Other fields here
    zip = USZipCodeField(blank=True, null=True)
```

Form uses a simple `ModelForm`

Error:
```
'NoneType' object has no attribute 'strip'
Request Method:	POST
Request URL:	http://localhost:8000/REDACTED
Django Version:	1.11.2
Exception Type:	AttributeError
Exception Value:	
'NoneType' object has no attribute 'strip'
Exception Location:	/REDACTED/localflavor/us/forms.py in to_python, line 45
```

Downgrading to Django 1.10.x fixes the issue. Looks like older versions of Django return a '' empty string if blank in which you can call `strip()` on it but Django 1.11.x returns a None in which you cannot strip.

peterfarrell:
Fixes form validation crash for USZipCodeField when field is null=True, blank=True.  See #294 

Do not call strip on a None.

**All Changes**

      already there.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -73,6 +73,7 @@ Authors
 * Olivier Sels
 * Paul Donohue
 * Paulo Poiati
+* Peter J. Farrell
 * Rael Max
 * Ramiro Morales
 * Rolf Erik Lekang
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,7 +18,8 @@ Modifications to existing flavors:
 
 Other changes:
 
-- None
+- Fixed crash with USZipCodeField form validation when null=True is allowed
+  (`gh-295 <https://github.com/django/django-localflavor/pull/295>`_)
 
 1.5   (2017-05-26)
 ------------------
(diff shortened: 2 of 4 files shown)


benkonrath:
Thanks for the fix. Can you add a failing test that's fixed by this change? It should be enough to test the form field with `empty_value` set to `None` instead of the default empty string.
https://docs.djangoproject.com/en/1.11/ref/forms/fields/#charfield

peterfarrell:
@benkonrath Added a test. It's a little hackish because Django < 1.11 don't support passing in `empty_value` to the field itself as it throws an unexpected argument error so instead of banging my head against the wall I went with the route that is in the test. Eventually it can be removed after support for Django < 1.11 happens.

Is there anything else you need to get this pulled in?

benkonrath:
Seems like a good solution. I'll make a 1.5.1 release with this fix soon.

PR #297, opened 2017-06-26
Fixed IBANFormField to support empty iban

TomRaz:
Currently, when using IBANField in django admin, and the field is nullable and blank=true, an exception is raised when trying to save the form without an iban.
I added handling of this case in the IBANFormField.
**All Changes**

- [V ] Add an entry to the docs/changelog.rst describing the change.

- [ V] Add an entry for your name in the docs/authors.rst file if it's not
      already there.

- [V ] Adjust your imports to a standard form by running this command:

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -92,3 +92,4 @@ Authors
 * Tom Forbes
 * Venelin Stoykov
 * Vladimir Nani
+* Tom Raz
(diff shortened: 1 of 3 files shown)


benkonrath:
Thanks for pointing out this issue and for the fix. It seems it's a general problem for a lot of the form fields that inherit `forms.CharField`. I'm working on a complete fix for all of the fields in #298. Let's keep this open until #298 is merged.

PR #298, opened 2017-06-28
Empty value django 1.11

benkonrath:
This PR adds support for the `empty_value` kwarg that was added to `forms.CharField` in Django 1.11.

On Django < 1.11, form fields that inherit from `forms.CharField` set an empty string for values that are in `self.empty_values`. On Django >= 1.11, form fields that inherit from `forms.CharField` set `self.empty_value` for values that are in `self.empty_values`. This means all of the fields now support the `empty_value` kwarg even when it's `None`.

The downside to this change is that form fields that inherit from `forms.CharField` will need to use the `EmptyValueCompatMixin` until we drop support for Django < 1.11.

Related #294 #295
Closes #297

### DIFF
--- a/localflavor/ar/forms.py
+++ b/localflavor/ar/forms.py
@@ -3,10 +3,11 @@
 
 from __future__ import unicode_literals
 
-from django.core.validators import EMPTY_VALUES
 from django.forms import ValidationError
 from django.forms.fields import CharField, RegexField, Select
 from django.utils.translation import ugettext_lazy as _
+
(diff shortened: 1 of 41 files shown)


benkonrath:
This is ready to go but I'd like to get some feedback if somebody has some time to double it. I'm planning to cherry pick 7d7356dd16b69512de054313790554d4f08b3c77 into the 1.5.x branch and make a 1.5.2 release

claudep:
Looks good to me, thanks!

PR #303, opened 2017-07-18
Fix COFA validation for USStateField

elliottcf:
The USPS codes for COFA states aren't in the `STATES_NORMALIZED` dictionary so they won't validate correctly in a `USStateField` form field.

I should have caught this in my previous PR, sorry.


benkonrath:
Is it possible or does it make sense to add some sort of test?

elliottcf:
I'm not against it. If it checked that all fields in the dropdown validate, it would have caught this bug.

benkonrath:
That sounds good. I'll cherry-pick this change to the 1.5.x branch as well.

benkonrath:
@elliottcf Are you still interested in contributing this fix? If so, it would be great to get a test for this change so that I can merge it in. Thanks.

elliottcf:
Yes, I am. I've just been busy with other things so I haven't gotten around to working on the tests you suggested.
I will try to get those written soon.

benkonrath:
No problem, I know the feeling. :-) It would be nice to get this into the 2.0 release which I'm hoping to release before the end of the year (see #322). I can always add this fix for 2.0.x and 1.6.x if it doesn't make it into 2.0. Thanks for your help with this.

Issue #258, opened 2016-11-06
Depricate non pep8 compliant class and method names


unloder:
**class SGNRIC_FINField(CharField):**
The name of this field could not be changed for backwards compatibility reasons. This field should get renamed and eventually deprecated when a deprecation solution will get implemented.
localflavor/sg/forms.py
  Line: 62
    pep8: N801 / class names should use CapWords convention (col 7)

**def DV_maker(v):**
/home/unloder/WWW/django-localflavor/localflavor/br/forms.py
    line 100

**mccabe: MC0001 / IDLicensePlateField.clean is too complex (12)**
localflavor/id_/forms.py
Line: 111

**mccabe: MC0001 / NOSocialSecurityNumber.clean is too complex (13)**
localflavor/no/forms.py
Line: 46

**pep8: E501 / line too long (137 > 120 characters) (col 121)**
localflavor/us/us_states.py
Line: 153

benkonrath:
These issues are all fixed.

PR #305, opened 2017-07-20
Deprecations and cleanups

benkonrath:
* More work on depreciations (see #258).
* Drop support for Django 1.9
* Small cleanup for IBAN tests.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -15,12 +15,14 @@ New fields for existing flavors:
 
 Modifications to existing flavors:
 
-- None
+- Fixed crash with USZipCodeField form validation when null=True is allowed
+  (`gh-295 <https://github.com/django/django-localflavor/pull/295>`_).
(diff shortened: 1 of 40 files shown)


claudep:
Fine, but do we gain anything by removing 1.9 support? If not, I wonder if it's worth it.

benkonrath:
There's no specific gain for dropping support for 1.9 except being explicit about not supporting version of Django that are no longer supported. I'll remove this change for now.

claudep:
I see your point. But as we generally gain nothing technically when dropping "intermediate" unsupported releases, my opinion is that we should keep that support. But this is in no way a final decision and this might be the subject of further discussions (e.g. on django-developers).

benkonrath:
Yeah, I understand it's just your opinion and I also see your point. For me it's not really a big deal either way - I dropped 1.9 support to save me a bit of time waiting for CI test and because I don't want to fix any bugs in old versions of Django. But can just decide to drop support if or when 1.9 (or any other unsupported version) is holding us back like we have in the past.

PR #306, opened 2017-07-26
Travis - test on Trusty

adamchainz:
As per [their blog post](https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming) they're making it the new default, best to be ahead of the curve.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 sudo: no
+dist: trusty
 language: python
 cache: pip
 


benkonrath:
I personally think it would be better to just leave it as the default since we're not using any OS specific features. This saves us having update this setting again in the future.

adamchainz:
Well at least this proves it's passing right now. I already had commits merged on many other django repos.

benkonrath:
The fact that a similar change has been merged in other Django repos isn't relevant unless Django is standardizing the Travis CI configs in some way. This change isn't needed for localflavor because we don't use any system level or OS specific features in Travis. Adding a change like this just adds noise and something to maintain that we don't need.

I hope that I have convinced you that localflavor doesn't need this update to the Travis config and that you'll close the PR. I'm specifically not closing the PR in case other locaflavor maintainers disagree with me and want to merge this change. At the end of the day it's a pretty minor point if goes either way. :-)

adamchainz:
At least we know it's not going to change anything

claudep:
Ben is the main localflavor maintainer and I respect his choice.

PR #307, opened 2017-07-28
Added deprecation warning to NLBankAccountNumberField

jieter:
ref #150 

**All Changes**

      already there.

### DIFF
--- a/localflavor/nl/models.py
+++ b/localflavor/nl/models.py
@@ -3,7 +3,7 @@ from __future__ import unicode_literals
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 
-from localflavor.deprecation import DeprecatedPhoneNumberField, RemovedInLocalflavor20Warning
+from localflavor.deprecation import DeprecatedPhoneNumberField
 
 from . import forms
 from .nl_provinces import PROVINCE_CHOICES
--- a/localflavor/nl/models.py
+++ b/localflavor/nl/models.py
@@ -1,9 +1,11 @@
 from __future__ import unicode_literals
 
+import warnings
+
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 
-from localflavor.deprecation import DeprecatedPhoneNumberField
+from localflavor.deprecation import DeprecatedPhoneNumberField, RemovedInLocalflavor20Warning
 
 from . import forms
 from .nl_provinces import PROVINCE_CHOICES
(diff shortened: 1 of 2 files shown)


localflavor/nl/models.py
```diff
@@ -117,9 +119,15 @@ class NLBankAccountNumberField(models.CharField):
...
     .. versionadded:: 1.1
+
+    .. deprecated:: 1.6
+        Use `localflavor.generic.models.IBANField` with included_countries=('nl') option instead.
     """
 
     def __init__(self, *args, **kwargs):
+        warnings.warn("NLBankAccountNumberField is deprecated. Use `localflavor.generic.models.IBANField` "
+                      "with included_countries=('nl') option instead.", RemovedInLocalflavor20Warning)
+
```
benkonrath: For deprecating the phone number models we're using the system check framework. Do you think it make sense to use the system check framework here as well?
https://github.com/django/django-localflavor/blob/master/localflavor/deprecation.py#L21-L24

Also, `IBANField` isn't a direct replacement for `NLBankAccountNumberField`. Users would need to do a data migration and generate the IBAN for each NL bank account number. My feeling is that adding this data migration is out of scope for localflavor. Calculating the check digits straightforward but we'd need to somehow map the NL account numbers to a 4 character IBAN bank ID and I don't this mapping readily available. I think we should just have a note stating that a data migration is required if users want to directly replace the  `NLBankAccountNumberField` with the `IBANFeild`. Otherwise, user will need to use both fields in parallel for a while.

jieter:
@benkonrath, holidays came in the way, I'll try to work on this coming week.

localflavor/nl/models.py
jieter: I have no preference regarding the systems check framework vs using `warnings.warn`. If you do prefer using the systems check framework, I'll be happy to change the patch.

You're right about the data migration, I'll add a note.

benkonrath: I prefer the systems check framework to keep things consistent with the other model deprecations. Thanks.

jieter:
@benkonrath  changed the deprecation warning and also rebased on current master.

benkonrath:
Looks good. You just need to remove the unused import.

```
localflavor/nl/models.py
  Line: 6
    pyflakes: F401 / 'localflavor.deprecation.RemovedInLocalflavor20Warning' imported but unused (col 1)
```

PR #311, opened 2017-10-30
Fixes NOBankAccountNumber issue with formatting and uncleaned data

thor:
If you would input accounts that were too short or otherwise riddled
with spaces and periods, it would not remove these before returning them again.

Fixes the issue and adds a couple of assertions.

I would :heart: it if this could come before 1.6 though, @benkonrath. 👍 

**All Changes**

      already there.

- [X] Adjust your imports to a standard form by running this command:

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -31,6 +31,8 @@ Other changes:
 - Added support for empty_value kwarg in Django >= 1.11
   (`gh-298 <https://github.com/django/django-localflavor/pull/298>`_).
 - Dropped support for Python 3.2.
+- Fixes issue with `no.forms.NOBankAccountNumber` unclean data.
+  (`gh-311 <https://github.com/django/django-localflavor/pull/311>`_).
 
 1.5   (2017-05-26)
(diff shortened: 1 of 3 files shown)


benkonrath:
Thanks. Does it make sense to put out a bug fix release for 1.5 with this change?

thor:
I would say yes, if you could, that would be most appreciated. There's no
backwards compatibility at all being changed here, it's "simply" (dangerous
word) fixing the `prepare_value()` for it so that it's not messing a user
about when correcting the inputted value.

I've got a project using my fork of master just because I discovered this
(to my great annoyance of myself).

If you could push out a small release, if 1.6 is a bit further ahead with
the backwards incompatible changes, that'd be lovely.

4. nov. 2017 23:33 skrev "Ben Konrath" <notifications@github.com>:

Thanks. Does it make sense to put out a bug fix release for 1.5 with this
change?

benkonrath:
I just made the 1.5.3 release with this fix. Thanks again for your contribution.

PR #316, opened 2017-11-22
Fix Czech birth number validation. Fixes #315

### LINKED ISSUE #315: Czech birth number validation issue
As stated in the source code ([here](https://github.com/django/django-localflavor/blob/master/localflavor/cz/forms.py#L74))

> Three digits for verification number were used until 1. january 1954

And then there is the following condition check:

```python
    if len(id) == 3:
        return '%s' % value
```

This means, that birth numbers that were issued after 1954.1.1 but don't have fourth digit in it will not be validated. But in fact they are invalid. For example:

**Birth number**: 600110/0006

This one passes the validation, but if I remove the last digit it will still pass the validation, but it should not because it has to have control digit.

PR: #316

Stranger6667:
As I described the issue in #315 these changes fix validation for birth numbers, that are issued after 1st of January 1954. All numbers after that date should have control digit.

**All Changes**

      already there.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -18,7 +18,8 @@ Modifications to existing flavors:
 
 Other changes:
 
-- None
+- Fixed validation of Czech birth numbers for birth dates after 1st January 1954
+  (`gh-315 <https://github.com/django/django-localflavor/issues/315>`_).
 
 1.6   (2017-11-22)
 ------------------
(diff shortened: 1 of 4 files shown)


docs/changelog.rst
```diff
@@ -18,7 +18,8 @@ Modifications to existing flavors:
 
 Other changes:
 
-- None
+- Fixed validation of Czech birth numbers for birth dates after 1st January 1954
+  (`gh-315 <https://github.com/django/django-localflavor/issues/315>`_).
```
benkonrath: Can you put this in the `Modifications to existing flavors` section? Thanks.

Stranger6667: Done

PR #324, opened 2017-12-30
Fix #210 -- Added VAT identifcation number validator

### LINKED ISSUE #210: EU VAT numbers
Hi I'm maintaining `django-vies` a package for european VAT number verification and validation.

I don't use the package myself and like to give it await. Since we are on djangocon it was suggested to merge it into localflavor.

I just did a complete rewrite of the package because I wanted it to be based on a validator rather than on fields only.

It's currently translate in like 3 languages, I'd like to improve that as well. I guess with the help of you guys it should be much easier.

I think there are a couple of different options:
1. You don't want it (which is ok)
2. You only want the validator part, not the model and form fields or widgets
3. You like it all.

I'd like to explain something, that might effect this decision:
The verification is based on regex, which can happen in a validator and works great.
The validation on the other hand is based on the VIES API. Which is great, because it tells you if a number is actually registered but it is not very reliable and should not be executed within a request.

Anyways:
The new code is [HERE](https://github.com/codingjoe/django-vies/tree/version-3)
Feel free to ask me anything, I'm looking forward to here your feedback.

Cheers from DjangoCon EU
Joe

codingjoe:


benkonrath:
I'll be able to look at this in a couple of days.

codingjoe:
@benkonrath I added the author and change log entry.

PR #327, opened 2018-02-19
Add NLLicensePlateField with tests

jarmovanlenthe:
Added a ModelField and Validator for Dutch License Plates (`NLLicensePlateField`) and a FormField (`NLLicensePlateFormField`) which also sanitizes input.

**All Changes**

- [DONE] Add an entry to the docs/changelog.rst describing the change.

- [DONE] Add an entry for your name in the docs/authors.rst file if it's not
      already there.

- [DONE] Adjust your imports to a standard form by running this command:

**New Fields Only**

- [DONE] Prefix the country code to all fields.

- [DONE] Field names should be easily understood by developers from the target
      tax / VAT ID etc.

- [DONE] Prefer '<country code>PostalCodeField' for postal codes as it's
      States postal system.

- [DONE] Add meaningful tests. 100% test coverage is not required but all

- [DONE] Add `.. versionadded:: <next-version>` comment markers to new
      localflavors.

- [DONE] Add documentation for all fields.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,7 +14,7 @@ New fields for existing flavors:
 
 Modifications to existing flavors:
 
-- Added `NLLicensePlateField` to NL flavor.
+- None
 
 Other changes:
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -10,11 +10,11 @@ New flavors:
 
 New fields for existing flavors:
 
-- None
+- `NLLicensePlateField` in NL flavor.
 
 Modifications to existing flavors:
 
-- None
+- Added `NLLicensePlateField` to NL flavor.
 
 Other changes:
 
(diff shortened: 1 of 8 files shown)


docs/changelog.rst
```diff
@@ -10,11 +10,11 @@ New flavors:
 
 New fields for existing flavors:
 
-- None
+- `NLLicensePlateField` in NL flavor.
 
 Modifications to existing flavors:
 
-- None
+- Added `NLLicensePlateField` to NL flavor.
```
benkonrath: The entry is only needed once in the 'New fields for existing flavors' section.

localflavor/nl/models.py
```diff
@@ -82,3 +82,28 @@ def formfield(self, **kwargs):
         defaults = {'form_class': forms.NLBSNFormField}
         defaults.update(kwargs)
         return super(NLBSNField, self).formfield(**defaults)
+
+
+class NLLicensePlateField(models.CharField):
+    """
+    A Dutch car license plate.
```
benkonrath: It's probably better to remove 'car' as this can be used for all Dutch license plates.

Issue #331, opened 2018-04-16
EAN validator is too picky


OdyX:
In the context of the https://github.com/defivelo/db project, I have two EAN numbers which validate on https://www.getnewidentity.com/validate-ean.php but are not considered valid by from `localflavor.generic.validators.EANValidator`:

```
>>> from localflavor.generic.validators import EANValidator
>>> v = EANValidator()
>>> v('7567554394380')
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "…/env/local/lib/python2.7/site-packages/localflavor/generic/validators.py", line 272, in __call__
    raise ValidationError(self.message, code='invalid')
ValidationError: [u'Not a valid EAN code.']
>>> >>> v2("756.4334.3999.00")
Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "…/env/local/lib/python2.7/site-packages/localflavor/generic/validators.py", line 272, in __call__
    raise ValidationError(self.message, code='invalid')
ValidationError: [u'Not a valid EAN code.']
```

These don't work for instance:
* `756.7554.3943.80`
* `756.4334.3999.00`

As mentionned in the pull request, this is an issue about the `calculated_checksum` which can be `10`, which cannot be compared to a single-digit.

PR #332, opened 2018-04-16
Fix EAN validation when intermediate checksum is 10

OdyX:
It was buggy because:
* `calculated_checksum % 10` could be `0` if `calculated_checksum` is a multiple of 10
* `10 - 0 = 10`
* `10` cannot be validly string-compared to a single-digit

Fixes: #331

- [x] Add an entry for your name in the docs/authors.rst file if it's not already there.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -29,6 +29,7 @@ Authors
 * Daniela Ponader
 * Danielle Madeley
 * Daniel Roschka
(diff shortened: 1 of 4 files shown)


benkonrath:
Thanks for the fix!

OdyX:
Yay, thank you very much for the very quick merge!

PR #337, opened 2018-06-01
Add field for greek Social Security Number (AMKA)

spapas:
- [x ] Add an entry to the docs/changelog.rst describing the change.

- [Was there ] Add an entry for your name in the docs/authors.rst file if it's not
      already there.

- [x ] Adjust your imports to a standard form by running this command:

### DIFF
--- a/localflavor/gr/forms.py
+++ b/localflavor/gr/forms.py
@@ -105,9 +105,9 @@ class GRSocialSecurityNumberCodeField(Field):
         val = re.sub('[\-\s\(\)]', '', force_text(value))
         if(len(val) < 11):
             raise ValidationError(self.error_messages['invalid'])
+        if self.allow_test_value and val == '00000000000':
+            return val
         if not all(char.isdigit() for char in val):
-            raise ValidationError(self.error_messages['invalid'])
-        if not self.allow_test_value and val == '00000000000':
             raise ValidationError(self.error_messages['invalid'])
 
         self.check_date(val)
--- a/localflavor/gr/forms.py
+++ b/localflavor/gr/forms.py
@@ -6,6 +6,7 @@ from django.core.validators import EMPTY_VALUES
 from django.forms import Field, RegexField, ValidationError
 from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
+from localflavor.generic.checksums import luhn
 
 
 class GRPostalCodeField(RegexField):
@@ -63,7 +64,7 @@ class GRTaxNumberCodeField(Field):
         return val
 
 
-class GRSocialSecurityNumberCodeField(Field):
+class GRSocialSecurityNumberCodeField(RegexField):
     """
     Greek social security number (AMKA) field.
 
@@ -77,7 +78,7 @@ class GRSocialSecurityNumberCodeField(Field):
 
     def __init__(self, allow_test_value=False, *args, **kwargs):
         self.allow_test_value = allow_test_value
-        super(GRSocialSecurityNumberCodeField, self).__init__(*args, **kwargs)
+        super(GRSocialSecurityNumberCodeField, self).__init__(r'^[0-9\s\-]+$', *args, **kwargs)
 
     def check_date(self, val):
         try:
@@ -85,25 +86,12 @@ class GRSocialSecurityNumberCodeField(Field):
         except:
             raise ValidationError(self.error_messages['invalid'])
 
-    def check_lunh(self, val):
-        # Uses the Luhn algorithm: https://en.wikipedia.org/wiki/Luhn_algorithm
-        s = 0
-        for idx, d in enumerate(reversed(list(map(int, val)))):
-            if (idx + 1) % 2 == 0:
-                d *= 2
-            if d > 9:
-                d -= 9
-            s += d
-        if s % 10 != 0:
-            raise ValidationError(self.error_messages['invalid'])
-
     def clean(self, value):
         super(GRSocialSecurityNumberCodeField, self).clean(value)
-        if value in EMPTY_VALUES:
-            return ''
-
-        val = re.sub('[\-\s\(\)]', '', force_text(value))
-        if(len(val) < 11):
+        if value in self.empty_values:
+            return self.empty_value
+        val = re.sub('[\-\s]', '', force_text(value))
+        if not val or len(val) < 11:
             raise ValidationError(self.error_messages['invalid'])
         if self.allow_test_value and val == '00000000000':
             return val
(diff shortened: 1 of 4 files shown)


spapas:
Finally, this PR should be ready for merge! 

As a general notice, submitting code to this repository is *way* to difficult, it took me like 50 times more time to fix the various satellite things (submission changes, prospector warnings, codecov etc) than to write the *actual* code for the new FormField! 

I recommend opening an issue for discussing this problem a bit...

localflavor/gr/forms.py
```diff
@@ -60,3 +61,56 @@ def clean(self, value):
         if mod != check:
             raise ValidationError(self.error_messages['invalid'])
         return val
+
+
+class GRSocialSecurityNumberCodeField(Field):
```
benkonrath: It would be better use `RegexField` instead of `Field`. You would then set the regex on the call to super.

localflavor/gr/forms.py
```diff
@@ -60,3 +61,56 @@ def clean(self, value):
...
+        self.allow_test_value = allow_test_value
+        super(GRSocialSecurityNumberCodeField, self).__init__(*args, **kwargs)
+
+    def check_date(self, val):
+        try:
+            datetime.datetime.strptime(val[:6], '%d%m%y')
+        except:
+            raise ValidationError(self.error_messages['invalid'])
+
+    def check_lunh(self, val):
```
benkonrath: Please use the `lunh` function that's already in localflavor instead of including your own version.

https://github.com/django/django-localflavor/blob/d23320486862d993070de51ef92e31b8f4cc61a3/localflavor/generic/checksums.py#L10

localflavor/gr/forms.py
```diff
@@ -60,3 +61,56 @@ def clean(self, value):
...
+            if d > 9:
+                d -= 9
+            s += d
+        if s % 10 != 0:
+            raise ValidationError(self.error_messages['invalid'])
+
+    def clean(self, value):
+        super(GRSocialSecurityNumberCodeField, self).clean(value)
+        if value in EMPTY_VALUES:
+            return ''
```
benkonrath: I think you won't need this check if you inherit from `RegexField` since there are checks for empty values in `CharField`.

localflavor/locale/el/LC_MESSAGES/django.po
```diff
@@ -14,7827 +14,10656 @@ msgid ""
 msgstr ""
```
benkonrath: Please remove the changes to this file. The localflavors take care of updating the translation files (which are managed through Transifex).

localflavor/gr/forms.py
spapas: Ok

localflavor/gr/forms.py
spapas: Ok

localflavor/gr/forms.py
spapas: Actually this:

```
        if value in self.empty_values:
            return self.empty_value

```
was needed or else the test ``test_forms_char_field_empty_value_allows_none`` would fail.

localflavor/locale/el/LC_MESSAGES/django.po
spapas: Ok I restored it to the previous version

spapas:
Hello @benkonrath, I've implemented the requested changes, I hope the PR will get merged soon!

Concerning my rant about the work needed for the PR, I understand that you want to have some standards for contributed code, however think about how much work is needed for guys that don't do this every day. You probably have setup your environments, tox, isort, prospector etc so that you can write code that passes all the tests automatically however most people don't have that comfort. 

I consider myself a rather experienced developer and setting up all the things needed to make the contribution and fixing things up was a major PITA; if I knew that it needed so much work just to add a form field (which I already use in my project) then I probably wouldn't bother at all. I had contributed some other fields before some years when each localflavor package was separate (https://github.com/spapas/django-localflavor-gr) the process back then was very easy. 

In any case, concerning the nature of the project (localflavor needs contributions for many people across the globe; you can't substitute quality for quantity here) I think that it does not make sense to have the same contribution standards as with Django. You need to lower your standards to encourage more contributions. After all, most of these contributions are values lists and form fields. The people that are maintaining this project can pick the individual contributions and make them as good as they want; however you don't want to discourage from contributing by making them jump hoops.

I hope you get the idea about my initial rant, really sorry if I ranted again trying to explain my thoughs!

Best regards and thanks for your work,
Serafeim

localflavor/gr/forms.py
benkonrath: Ok cool. Thanks for catching that.

benkonrath:
@spapas Thanks again for your contribution.

I see your point about making it easier to contribute to localflavor. I think we just have a different value on what's a suitable quality level for this project. Perhaps we could improve the documentation on how to test locally rather than having to wait for Travis to fail. Suggestions or PRs for this would be helpful.

I'm not the last word on this. I'm only the most active localflavor maintainer at the moment. If you want to discuss this further, perhaps you could open a new issue about this like you suggested. Just a warning though, I don't know if you'll get much of a response because the other maintainers aren't too active these days. People who watch the repo can also comment here if they have an opinion about this.

claudep:
It's not easy to draw the balance between code quality and ease of contribution. For me, testing/coverage is a must and we shouldn't lower that quality requirement. It's one of the Django mantras: "untested code is broken code".
I'm not so much convinced about other style-related requirements. That stuff could also be handled by the committer at a final stage. On the other hand, it may encourage devs to adopt good coding habits.

spapas:
Hello @benkonrath thank you very much for merging my PR!

Now, for the quality/easy of contribution, I understand your concerns and am all in for code quality. However I agree with @claudep that the bar for contributions is too high for most developers and, due to the nature of the project as explained in my previous comment, it would be better to lower it a bit to encourage more contributions (and merging existing ones). I believe that being able to find some specific form fields for your country is a factor that weights when you are a lone developer in (f.e) Greece and want to decide which framework you'll want to learn!

In any case, what was in my mind would be to allow and quickly merge even non perfect PRs that would then be fixed/improved (by adding more tests, complying with PEP8 etc) by the original contributor or the maintainers; you'll add an issue with what is missing and then assign it to the original contributor; if he doesn't fix it after a while then a maintainer will step up and fix it for him.

However after Ben's previous comment (that most maintainers are not so active) I'm not sure if that's possible because it will add a lot of effort to the maintainers (to add the corresponding issue and fix it if needed) so probably the contribution workflow should stay as it for now :/

Issue #334, opened 2018-04-26
Model fields to brazillian documents


andrelramos:
I'm working on a system that needs some specific brazilian fields. Some of them already has in br form fields but a have many model forms that needs this fields and it would be nice if they were created as model fields. 

The fields are:

- Brazilian documents (CPF and CNPJ)
- Brazilian Telephone and Cell Phone format
- Brazilian Zip Code

In 2015 i was started a PR with code to solve it (#188) but it was denied, so i'm creating this issue to check the idea.

benkonrath:
It's fine by me. I don't know what the PR was turned down last time. I'll review this PR when you make it. 

It's best to extract the validation into a separate validator and use this validator for the form and the model. You can look at some the example models already included in localflavor to see how it's done (e.g. the IBAN fields).

If you're not sure about something, just make a 'work in progress' PR and we can discuss things on the PR. Thanks!

benkonrath:
Fixed in 9d47350d9e4f76d3a24e66750281ad8a380da76e.

PR #341, opened 2018-07-28
Add Brazilian Model Fields

andrelramos:
Closes https://github.com/django/django-localflavor/issues/334


localflavor/br/validators.py
```diff
@@ -0,0 +1,48 @@
...
+    """A validator for Brazilian Zip Codes (CEP)."""
+
+    def __init__(self, *args, **kwargs):
+        self.regex = '^[0-9]{5}[-]?[0-9]{3}$'
+        self.message = _('Enter a zip code in the format 00000000 or 00000-000.')
+        self.code = _('Invalid Zip Code')
+        super(BRZipCodeValidator, self).__init__(*args, **kwargs)
+
+
+class BRCNPJValidator(RegexValidator):
```
benkonrath: Rather than using a simple reqex validator, the validation logic from the `BRCNPJField` form field needs to be extracted and put here. You'd then use this validator in both the form and model field. This way we won't be including two validation algorithms for the same data type.


The same applies to the `BRCPFValidator` and the `BRCPFField` form field (below).

localflavor/br/models.py
```diff
@@ -18,3 +22,68 @@ def deconstruct(self):
...
+
+    def __init__(self, *args, **kwargs):
+        kwargs['max_length'] = 14
+        super(BRCPFField, self).__init__(*args, **kwargs)
+        self.validators.append(validators.BRCPFValidator())
+
+
+class BRCNPJField(CharField):
+    """
+    A model field for the brazilian document named of CNPJ (Cadastro Nacional de Pessoa Jurídica)
```
benkonrath: Please add `.. versionadded:: 2.1` to the new fields. Hopefully we can get this into the next release. :-)

andrelramos:
@benkonrath Thank you for your suggestions!

I finished the requested changes. Can you take a look again?

I'm very glad to know that these changes will be introduced in next release. :grinning:

localflavor/br/forms.py
```diff
@@ -137,7 +109,7 @@ class BRCNPJField(CharField):
     Otherwise both formats will be valid.
 
     .. _Brazilian CNPJ: http://en.wikipedia.org/wiki/National_identification_number#Brazil
-    .. versionchanged:: 1.4
+    .. versionchanged:: 2.1
```
benkonrath: You shouldn't remove the old version changed marker because it's useful for people to know the history. I realize that the 1.4 directive is missing the explanation but it's still helpful because people can check the changelog if they need more information about the change.

localflavor/br/forms.py
```diff
@@ -147,34 +119,15 @@ class BRCNPJField(CharField):
...
-        value = value[:-2] + str(new_1dv) + value[-1]
-        new_2dv = sum([i * int(value[idx]) for idx, i in enumerate(list(range(6, 1, -1)) + list(range(9, 1, -1)))])
-        new_2dv = dv_maker(new_2dv % 11)
-        value = value[:-1] + str(new_2dv)
-        if value[-2:] != orig_dv:
-            raise ValidationError(self.error_messages['invalid'])
-
-        return orig_value
+        original_value = value[:]
+        return original_value
```
benkonrath: You should just do `return value` instead of these 2 lines. The same applies to the `BRCPFField`.

localflavor/br/forms.py
```diff
@@ -11,12 +11,10 @@
 from django.utils.encoding import force_text
 from django.utils.translation import ugettext_lazy as _
 
+from localflavor.br import validators
```
benkonrath: For consistency, can you use relative imports like this: `from .validators import postal_code_re, ...`. This applies to other files as well.

localflavor/br/forms.py
```diff
@@ -79,6 +71,8 @@ class BRCPFField(CharField):
 
     More information:
     http://en.wikipedia.org/wiki/Cadastro_de_Pessoas_F%C3%ADsicas
+
+    .. versionchanged:: 2.1
```
benkonrath: A short note about what changed is useful. Something like "Use BRCPFValidator for validation." would work but feel free to write something yourself. The same applies to other `versionchanged` directives.

http://www.sphinx-doc.org/en/stable/markup/para.html#directive-versionchanged

localflavor/br/models.py
```diff
@@ -18,3 +22,48 @@ def deconstruct(self):
...
+
+    .. versionadded:: 2.1
+    """
+
+    description = _("CNPJ Document")
+
+    def __init__(self, *args, **kwargs):
+        kwargs['max_length'] = 18
+        super(BRCNPJField, self).__init__(*args, **kwargs)
+        self.validators.append(validators.BRPostalCodeValidator())
```
benkonrath: I think this should be the `BRCNPJValidator`, not the `BRPostalCodeValidator`. I'm a bit surprised that the tests didn't catch this.

tests/test_br.py
```diff
@@ -177,3 +178,30 @@ def test_BRStateChoiceField(self):
             'pr': error_invalid,
         }
         self.assertFieldOutput(BRStateChoiceField, valid, invalid)
+
+
+class BRLocalFlavorModelTests(SimpleTestCase):
```
benkonrath: You should probably also check that the correct validator is set on the model fields just to make sure it's the correct one.

benkonrath:
We'll have to push this to 2.2.

benkonrath:
localflavor isn't tied to using the same release number as Django. I'd like to aim to make the 2.2 release in the first week of December (tentatively of course). This should be a good rallying point for this PR and other new features.

andrelramos:
@benkonrath I did the last changes request. I think that it's done now.

benkonrath:
I ended up rebasing and squashing the commit in #365 just to make it a little cleaner. Thanks again for your contribution and sorry for the delay in getting this merged.

PR #342, opened 2018-07-29
Add 34 prefix value (ar localflavor ARCUITField).

ramiro:
There are now CUIT assigned that use a 34 prefix. This PR adds support for it.

**All Changes**

      already there.

### DIFF
--- a/localflavor/ar/forms.py
+++ b/localflavor/ar/forms.py
@@ -11,8 +11,7 @@ from .ar_provinces import PROVINCE_CHOICES
 
 
 class ARProvinceSelect(Select):
-    """A Select widget that uses a list of Argentinean provinces/autonomous
-    cities as its choices."""
+    """A Select widget that uses a list of Argentinean provinces/autonomous cities as its choices."""
 
     def __init__(self, attrs=None):
         super(ARProvinceSelect, self).__init__(attrs, choices=PROVINCE_CHOICES)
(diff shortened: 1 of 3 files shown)


localflavor/ar/forms.py
```diff
@@ -11,7 +11,8 @@
 
 
 class ARProvinceSelect(Select):
-    """A Select widget that uses a list of Argentinean provinces/autonomous cities as its choices."""
+    """A Select widget that uses a list of Argentinean provinces/autonomous
+    cities as its choices."""
```
claudep: I think that we generally go to next line after `"""` when it takes more than one line, but that's nit-picking...

benkonrath: I agree. We're using 120 character lines so you can also just not include this change. Either way is fine.

localflavor/ar/forms.py
```diff
@@ -77,19 +79,19 @@ class ARCUITField(RegexField):
...
-    ACUIT is of the form XX-XXXXXXXX-V. The last digit is a check digit.
+    A CUIT is of the form XX-XXXXXXXX-V. The last digit is a check digit.
 
     More info:
     http://es.wikipedia.org/wiki/Clave_%C3%9Anica_de_Identificaci%C3%B3n_Tributaria
 
-    English info:
+    Info in English:
     http://www.justlanded.com/english/Argentina/Argentina-Guide/Visas-Permits/Other-Legal-Documents
     """
```
benkonrath: It would be great if you could add `.. versionchanged:: 2.1` here.

ramiro:
Thank you for your reviews! I think I've taken care of them.

PR #352, opened 2018-10-08
Update Indonesian localflavour

geomars:
Modification:
- [x] Update `PROVINCE_CHOICES` in **localflavor/id_/id_choices.py** to comply with official codes issued by Indonesian internal affair ministry
- [x] Update **test/test_id.py**

### DIFF
--- a/localflavor/id_/id_choices.py
+++ b/localflavor/id_/id_choices.py
@@ -1,42 +1,43 @@
 from django.utils.translation import ugettext_lazy as _
 
-#: Indonesia does not have an official Province code standard.
-#: I decided to use unambiguous and consistent (some are common) 3-letter codes.
-#: http://id.wikipedia.org/wiki/Daftar_provinsi_Indonesia
+# Provincial code in Indonesia doesn't comply with the standard 3-letter province code
+# data taken from the official page of Indonesian internal affair ministry
+# http://www.kemendagri.go.id/pages/data-wilayah
 PROVINCE_CHOICES = (
-    ('ACE', _('Aceh')),
-    ('BLI', _('Bali')),
-    ('BTN', _('Banten')),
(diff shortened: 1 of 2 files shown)


claudep:
Thanks, much better now! @benkonrath will probably review it in due time. But I don't think this will be accepted as is, because this would break existing usage for people having stored previous values in their models. I'm sure Ben will be able to help with coping with this sort of changes.

geomars:
Thanks for your review @claudep

benkonrath:
@geomars Thanks for this pull request and sorry for the delay in responding. Could you provide some background on why you want to change the codes that are stored for the provinces? Is it because the current codes aren't meaningful? 

Instead of changing the current version of `PROVINCE_CHOICES`, I think it's better to make a new version with a different name. You would also need to create new version of `IDProvinceSelect` that uses the new version of `PROVINCE_CHOICES`. Doing this will allow us to depreciate the current versions, highlight this deprecation in the changelog and eventually remove the field and the choices.

localflavor/id_/id_choices.py
```diff
@@ -1,42 +1,43 @@
 from django.utils.translation import ugettext_lazy as _
 
-#: Indonesia does not have an official Province code standard.
-#: I decided to use unambiguous and consistent (some are common) 3-letter codes.
-#: http://id.wikipedia.org/wiki/Daftar_provinsi_Indonesia
+# Provincial code in Indonesia doesn't comply with the standard 3-letter province code
+# data taken from the official page of Indonesian internal affair ministry
+# http://www.kemendagri.go.id/pages/data-wilayah
```
benkonrath: I get a 404 page not found error when I try to load this URL. Has this page been removed since you made this pull request?

PR #360, opened 2019-02-20
Ie model fields

paulcunnane:
This change adds an "EircodeField" to the Irish local flavour. An Eircode is
an Irish postcode (https://www.eircode.ie). The format of the Eircode is
validated, but its existence is not.

Thanks for your contribution!

first version.

**All Changes**

      already there.

**New Fields Only**

      I haven't done this, because "Eircode" is a uniquely Irish name for
      a postcode.

      tax / VAT ID etc.

      States postal system.

      Again, it's called "Eircode" because we just don't call them postal codes.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -16,10 +16,6 @@ first version.
 - [ ] Add an entry for your name in the docs/authors.rst file if it's not
       already there.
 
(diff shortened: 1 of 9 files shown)


benkonrath:
Thanks for your contribution. Can you rebase this branch with master? That will pickup the latest changes and fixes to the CI config and the tests should pass. I'll do the review once you've rebased. Thanks.

paulcunnane:
I think that's done now (not a git expert!)

PR #362, opened 2019-02-22
Remove isort from CI.

benkonrath:
This PR removes the isort check in the CI as suggested by a couple of users. I still like the consistency this brings but too many contributors are running into problems with this check so it's time to remove it. We can always run isort manually once in a while if we want to keep things consistent with the imports.

Any thoughts or opinion on this?


claudep:
I don't care much, feel free to make your decision.

benkonrath:
Ok, thanks for your input.

PR #366, opened 2019-03-23
ie model fields - rebase

benkonrath:
This PR contains the code by @paulcunnane in #360 rebased to master. I started doing a review but after my investigation I had 80% of the updates done so I thought I'd just finish it.

Changes:
* Convert to using `RegexField`
* Remove space from regex so that it accepts Eircodes without spaces
* Add a display format with a space
* Use the the built-in strip option

@paulcunnane Could you test this version out to make sure it's working well?

Closes #360

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -82,6 +82,7 @@ Authors
 * Morgane Alonso
 * Olivier Sels
 * Olle Vidner
+* Paul Cunnane
 * Paul Donohue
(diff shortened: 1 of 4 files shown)


benkonrath:
I assume this is working since the tests still pass after my changes. I'm going to merge this so that it's in the 2.2 release - we can always make changes if it needs to be adjusted for some reason.

PR #368, opened 2019-04-17
update python/django support: run tests for django 2.2

Niouby:
* run tests for Django 2.0 and Python 3.7
* run tests for Django 2.2 with the current released version instead of rc1


.travis.yml
```diff
@@ -1,12 +1,12 @@
 sudo: no
 language: python
 cache: pip
+dist: xenial
 
 matrix:
   include:
     # Need to use xenial in the docs build for a more recent version of sqlite.
```
claudep: Probably the comment can be removed now, and you also forgot some dist: xenial lines below.

Niouby: Thanks, I remove it. ^^

benkonrath:
Thanks for the PR!

Niouby:
thanks :+1:

PR #369, opened 2019-04-17
Removed Python 2 compatibility

claudep:
This removes Python 2 compatibility and makes django-localflavor run with next Django version (3.0).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,6 @@ matrix:
       env: TOXENV=docs
     - python: 3.5
       env: TOXENV=prospector
-    - python: 2.7
-      env: TOXENV=py27-1.11
-    - python: pypy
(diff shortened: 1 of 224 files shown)


benkonrath:
Thanks for this! I was roughly planning a 2.2 release at the end of the month. I think it's better to merge this after the 2.2 release so that we can support Django 1.11 & 2.2 in our 2.2 release. What do you think?

claudep:
Sure, that's a good plan!

claudep:
Just rebased the branch.

claudep:
See https://code.djangoproject.com/ticket/30399 about the failures we get on the master branch.

claudep:
The PR was committed, I triggered a new build and the tests now pass on master!

PR #370, opened 2019-04-18
Deprecate generic.checksums.luhn and ean

benkonrath:
I've been meaning to do this for a while. As the code suggests, my plan is to remove the `generic.checksums.luhn` and `generic.checksums.ean` functions in 3.0 - users should just use python-stdnum directly.

@claudep If you have time a review and your thoughts on this plan would be helpful. Thanks.


tests/test_deprecated.py
```diff
@@ -0,0 +1,22 @@
...
+from localflavor.generic.checksums import luhn
+
+try:
+    from unittest.mock import patch
+except ImportError:
+    from mock import patch
+
+
+class DeprecatedFieldsTests(SimpleTestCase):
+    @patch('localflavor.generic.checksums.stdnum_luhn')
```
claudep: Not sure, what is the role of patch here?

claudep: Oh, for the assertion below? All that just for testing a deprecation. I wouldn't have moaned if you omitted that :-)

tests/test_deprecated.py
```diff
@@ -0,0 +1,31 @@
...
+    def test_luhn_deprecated(self, stdnum_luhn):
+        with warnings.catch_warnings(record=True) as recorded:
+            warnings.simplefilter('always')
+            luhn('1234')
+
+        self.assertTrue(all(w.category is RemovedInLocalflavor30Warning for w in recorded))
+        stdnum_luhn.asert_called_once_with('1234')
```
claudep: Strange that the tests don't fail with the typ0 `asert` vs `assert`

benkonrath: That's mock - it silently allows undefined methods to be called without raising an error.

tests/test_deprecated.py
benkonrath: I agree it's a bit pedantic - thanks for pointing this out. Removing these tests also means I can remove the additional requirement for mock on Python 2 which simplifies things a bit.

tests/test_deprecated.py
benkonrath: By the way - nice catch! :-)

PR #374, opened 2019-05-11
Egypt local flavor

AhShXyZ:
**All Changes**

      already there.

**New Fields Only**

      tax / VAT ID etc.

      localflavors.


benkonrath:
Thanks for the PR! I'll have time to review this in the next few days. Thanks for your patience.

localflavor/eg/forms.py
```diff
@@ -0,0 +1,74 @@
...
+            date(int(year), int(month), int(day))
+        except ValueError:
+            raise ValidationError(self.error_messages['invalid'])
+
+        # is valid governorate code?
+        governorate_codes = (
+            '01', '02', '03', '04', '11', '12', '13', '14', '15', '16', '17', '18', '19', '21',
+            '22', '23', '24', '25', '26', '27', '28', '29', '31', '32', '33', '34', '35', '88')
+        if governorate_code not in governorate_codes:
+            raise ValidationError(self.error_messages['invalid'])
```
benkonrath: This invalid case isn't covered by the tests. Can you add an invalid id number to the tests that has an invalid governote code? - Just for completeness. Thanks.

AhShXyZ: Reached 98% code coverage instead of 93% hope that's the desired change. Thank you for your time.

benkonrath:

PR #378, opened 2019-06-15
Italian provinces updated: Sud Sardegna added and Carbonia-Iglesias, …

taifu:
Italian provinces updated:

- Sud Sardegna added
- Carbonia-Iglesias, Medio Campidano, Ogliastra, Olbia-Tempio removed

[2016 provinces changes](https://it.wikipedia.org/wiki/Province_d%27Italia_soppresse)

**All Changes**

      already there.


taifu:
> Thanks for the PR - it looks good. The only missing part is the change log entry. You should specifically mention in the change log entry that a data migration is required for users of the `ITRegionProvinceSelect`: `CI`, `VS`, `OG`, and `OT` will need to be converted to `SU`.

I don't think this is mandatory. For example an old customer should stay with his old province, without any data migration, whilst an active one should be migrated. What do you think? Am I wrong?

benkonrath:
Here's how I understand things: Django won't be able to generate a `selected` value for the choice field when one of the removed choices is stored in the database. In this case it will just show the default value, not the database value.

I'm guessing a bit here because it's been a while since I've encountered this behaviour. Are you able to do a little test in your local dev environment? You can just store one of the deleted values and see how it shows up in a form with this new version of the code.

taifu:
> Here's how I understand things: Django won't be able to generate a `selected` value for the choice field when one of the removed choices is stored in the database. In this case it will just show the default value, not the database value.

Yep. This is how it works. So what do you think is it right to write in the changelog?

benkonrath:
After rebasing (to get the latest update to the change log), you can make a normal change log entry about the change in the correct place and also add some information about the data migration above  'New flavors'.

For an example, you can see how I just updated the change log for removing the deprecated code:
https://github.com/django/django-localflavor/pull/379

PR #380, opened 2019-07-09
Extend Danish post code field with small Danish islands

HenrikOssipoff:
In 2017 a lot of small Danish islands gained their own independant postal codes, which were missing from  `DK_POSTCALCODES`. Also added 3 new service postal codes (ones below 1000).

**All Changes**

      already there.


HenrikOssipoff:
Reference Excel sheet from PostNord: https://www.postnord.dk/globalassets/danmark/.excel-docs/postnummerfil-til-download-06-19.xlsx

HenrikOssipoff:
@benkonrath Thank you for the comment, I've updated the PR as per your suggestion. Let me know if it's OK, and I'll squash everything.

benkonrath:
Looks good.

PR #381, opened 2019-07-18
Added cities of Israel and postal code regions in Great Britain

theBrownBug:
Added cities in Israel. 
Added postal code regions of Great Britain

Source : https://www.britannica.com/topic/list-of-cities-and-towns-in-Israel-2060226
https://en.wikipedia.org/wiki/List_of_postcode_areas_in_the_United_Kingdom

PR #382, opened 2019-07-18
Australian postal codes

theBrownBug:
Added postal codes for national capital area in Australia


benkonrath:
Thanks for your contribution. As Clauld mentioned, we will only accept PRs that are logically grouped together. So you have to split this PR into three separate PRs. I'm closing this PR and #381 but please do create new PRs for each change you'd like to contribute. Thanks.

PR #384, opened 2019-07-19
Added postal code areas in GB

theBrownBug:
Added Postal code areas of GB and its crown dependencies.

Source:
https://en.wikipedia.org/wiki/List_of_postcode_areas_in_the_United_Kingdom


leogregianin:
Sorry for the long delay in responding to this request.

This request has postal codes from Great Britain, it is a single file, without tests or any integration with existing classes.

In the class GBPostcodeField is already validation of the postcodes:
https://github.com/django/django-localflavor/blob/274c397189f182a3239b6c81540dbc98518210cb/localflavor/gb/forms.py#L12

PR #386, opened 2019-07-29
Change National Identity number validation for women

int2k:
Change test to use translated messages

**Please replace these instructions with a description of your change. The
'New Fields Only' section should be removed if your pull request
doesn't add any new fields.**

Thanks for your contribution!

first version.

**All Changes**

      already there.


benkonrath:
I noticed that this was in draft mode so I "published" it. Sorry for the delay in looking at this. I might have skipped it because of the draft mode. In any case, it looks good. Thanks for the PR!

PR #390, opened 2019-10-14
Updated test configs for Django 3.0b1.

benkonrath:


claudep:
Python 3.5 is not supported with Django 3.0

benkonrath:
Yep. I got it right in the tox config, just not the travis config ... too rushed at the end of the day I guess. :-/

PR #392, opened 2019-10-29
Use set value of strip in fields that inherit from forms.CharField.

benkonrath:
`django.forms.CharField` already strips the white space if requested. The default is `True` which is why it's safe to just remove the `strip()` in these form fields.
https://docs.djangoproject.com/en/2.2/ref/forms/fields/#django.forms.CharField.strip


claudep:
Nice!
I wouldn't add the `versionchanged` directives if the functionality is unchanged (which seems the case).

PR #394, opened 2019-11-27
Updated Icelandic post code field

HenrikOssipoff:
It seems like there were missing a *lot* of Icelandic post codes. It also appears that the list contained a lot of post codes that are actually not valid post codes.

The list have been updated based on the official data bank from Postur.is: https://www.postur.is/en/about-us/post-offices/post-codes/data-files/

I'm not sure if this is actually a valid change, given that this _removes_ some post codes? Should the new ones just be added, and the old ones kept?

Also note that some cities have seemingly changed names based on the newest data.

**All Changes**

      already there.


benkonrath:
Thanks for the PR and sorry for the delay. The next release will be 3.0 which can have breaking changes. The timing of this fix is good but we also have to consider this even if we're weren't doing a major version bump - we'd just have to include in a different way.

The PR looks good. The only addition would be specifically mentioning the breaking change as a new line under the notice for the removal of the checksum functions. Thanks.

HenrikOssipoff:
@benkonrath Thanks for the review. I've added a notice and re-worded the change log slightly. Let me know if it's good, and I'll squash the PR.

docs/changelog.rst
```diff
@@ -25,11 +27,14 @@ Modifications to existing flavors:
...
   `si.forms.SIEMSOField`, `si.forms.SITaxNumberField` and `za.forms.ZAIDField`
   (`gh-392 <https://github.com/django/django-localflavor/pull/392>`_).
+- Updated Icelandic `IS_POSTALCODES` with missing entries, updated spelling of entries, and removed non-existing ones.
+  (`gh-394 <https://github.com/django/django-localflavor/pull/394>`_).
 
 Other changes:
 
 - Removed deprecated `generic.checksums.luhn` and `generic.checksums.ean` functions
   (`gh-379 <https://github.com/django/django-localflavor/pull/379>`_).
+- Re
```
benkonrath: Was this included by mistake?

benkonrath: I've fixed this.

PR #401, opened 2020-02-13
Use the value returned by field's clean()

emesik:
I've noticed that for many flavors the `super().clean()` of fields is called but the result is ignored. Such behavior voids base class validation, for example the `strip=True` setting is always ignored.

This change unifies all flavors by using the results of base class validations.


benkonrath:
Looks good, thanks! Could you add a changelog entry for this? Thanks.

PR #406, opened 2020-02-16
Fix more deprecation warnings.

benkonrath:


localflavor/nz/forms.py
```diff
@@ -84,7 +84,7 @@ def clean(self, value):
         value = super().clean(value)
         if value in EMPTY_VALUES:
             return ''
-        value = re.sub('(\s+|-)', '', smart_str(value))
+        value = re.sub(r'(\s+|-)', '', smart_str(value))
```
claudep: Moreover, I don't think using the `smart` version add anything here, as we don't have any interest to keep an hypothetical lazy string.

benkonrath: Thanks, I've changed it.

PR #411, opened 2020-03-08
docs: Fix simple typo, validty -> validity

timgates42:
There is a small typo in localflavor/kw/forms.py.

Should read `validity` rather than `validty`.


benkonrath:
Thanks!

PR #412, opened 2020-03-08
Update prospector

benkonrath:
I'll be slowly working on this as I have extra time.


benkonrath:
@claudep Hi Claude, I think the changes here are ok but it would be nice if you could have a quick scan to double check this. Thanks!

claudep (approved):
Thanks Ben, this looks good. A further step would be to not allow raw excepts, I think we have only one currently in gr/forms.py, right?

benkonrath:
Yeah, just the one bare except that you pointed out. It's fixed now. Thanks for the review.

benkonrath:
I'm also fine with fixing the `no-else-return` and `no-else-raise` problems that are reported when these pylint options aren't disabled. I just wasn't sure how "wrong" or "bad" the "if return/else return" pattern really is. Do you have an opinion?

claudep:
I've just been recently confronted to that. I don't think it matters so much. Of course, having else after a return is unnecessary technically speaking, but in another view, having the else balancing the if does make some semantically sense to me. So do what you like.

benkonrath:
I also think the else makes sense semantically - although in new projects I usually just accept the pylint default and don't add the else when returning in a chain of if statements. I'll leave it as is since doesn't really cause too much harm. Thanks for your input on this.

PR #415, opened 2020-04-14
Fix French national identification number validation with people born overseas

leo-naeka:
I realized that localflavor does not allow the entry of NIR with `98` as the birth department, excluding it because `98` would correspond to Monaco.

In fact, `98` is used for "COM" (eg. French Polynesia, Wallis and Futuna, New Caledonia) and people born in Monaco would have `99` as `department_of_origin` and `138` as `commune_of_origin` (`99` stands for a birth abroad and `138` is the INSEE code for Monaco).

This codification is a real mess, I've attached two official INSEE documents, which are more complete and reliable sources than Wikipedia (page 6 is a good summary).

This fix add the following changes:
- Remove "Monaco's" invalidation
- Increase DOM dept range (`[971-976]` -> `[971-978]`)
- Support COM numbers (dept range within `[984-989]`)
- Add info for foreign cases (dept is `99`)
- Add according tests

---

[INSEE Guide identification.pdf](https://github.com/django/django-localflavor/files/4477931/INSEE.Guide.identification.pdf)
[INSEE Codes Pays.pdf](https://github.com/django/django-localflavor/files/4477930/INSEE.Codes.Pays.pdf)

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,7 +14,7 @@ New fields for existing flavors:
 
 Modifications to existing flavors:
 
-- None
+- Fix `FRNationalIdentificationNumber` validation for people born overseas
 
 Other changes:
 
--- a/localflavor/fr/forms.py
+++ b/localflavor/fr/forms.py
@@ -143,35 +143,41 @@ class FRNationalIdentificationNumber(CharField):
             raise ValidationError(self.error_messages['invalid'])
 
     def _clean_department_and_commune(self, commune_of_origin, current_year, department_of_origin, year_of_birth):
-        # Departments number 20, 2A and 2B represent Corsica
         if department_of_origin in ['20', '2A', '2B']:
-            # For people born before 1976, Corsica number was 20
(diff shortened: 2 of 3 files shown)


benkonrath:
It looks good, thanks!

Just two small remarks:

- It would be nice if you could include links to the PDFs that you attached. It's nice that they're attached on this issue but links to the reference documents in the code help make them a little more discoverable.

- You're welcome to add your name to the authors file although this is optional.

I can merge this as soon as the links are added if that's possible. Thanks.

leo-naeka:
> * It would be nice if you could include links to the PDFs that you attached. It's nice that they're attached on this issue but links to the reference documents in the code help make them a little more discoverable.

@benkonrath I had those documents for some time, grabbed from an administrative site but can't find their links anymore (and we definitively should not rely on these sites to preserve access over time), so I've put them on Scribd.

Ok for me to merge like this, let me know if you prefer these commits squashed.

benkonrath:
Thanks, I rebased and squashed the commits on merge (it's easy to do directly in github these days).

PR #422, opened 2020-10-02
Belarus localflavor

samgans:
Hi! I've added the localflavor for Belarus!

Released:

1. A number of new form fields (related to passports, regions, and postal codes); 
2. A number of new model fields with links to those of form fields;
3. Tests for proper validation;

### DIFF
--- a/localflavor/by/forms.py
+++ b/localflavor/by/forms.py
@@ -1,0 +1,102 @@
+from django import forms
+from django.utils.translation import ugettext_lazy as _
+
+from .by_regions import BY_REGIONS_CHOICES
+
+
+class BaseKwargsUpdatedField:
+    """
+    Abstract base class made to conform the DRY principle.
+
+    The initial_options is overrided by the dict in the subclasses and then
+    automatically passed to kwargs in the FormField.__init__() method.
+    """
+
+    initial_options = None
+
+    def __init__(self, *args, **kwargs):
+        kwargs.update(self.initial_options)
+        super().__init__(*args, **kwargs)
+
+
+class UpperValueMixin:
+    """
+    Mixin made to conform the DRY principle.
+
+    Overrides the to_python methods to return the value.upper()
+    """
+
+    def to_python(self, value):
+        value = super().to_python(value)
+        if value is None:
+            return value
+        else:
+            return value.upper()
+
+
+class BYRegionField(BaseKwargsUpdatedField, forms.TypedChoiceField):
+    """
+    A form field that ensures the input is a number representing the region.
+    """
+
+    initial_options = {
+        'choices': BY_REGIONS_CHOICES,
+        'empty_value': None
+    }
+
+
+class BYPassNumberField(BaseKwargsUpdatedField, UpperValueMixin, forms.RegexField):
+    """
+    A form field that validates its input is a pass number in a right format.
+
+    Apart from that, it tries to convert all characters to uppercase as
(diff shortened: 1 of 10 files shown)


benkonrath:
Thanks for your PR! I'll try to review it in the next couple of days.

localflavor/by/forms.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def to_python(self, value):
+        value = super().to_python(value)
+        if value is None:
+            return value
+        else:
+            return value.upper()
+
+
+class BYRegionField(BaseKwargsUpdatedField, forms.TypedChoiceField):
```
benkonrath, line 37: Instead of making `ChoiceField` for the regions, I think it's better to make a `forms.Select` widget that can be used in forms. The `BY_REGIONS_CHOICES` can be set directly on the model instead of using this custom form field.

Here's an example of a `forms.Select` widget:
https://github.com/django/django-localflavor/blob/456dd7e44537a3170c5d672dd51481885b0ff5e8/localflavor/md/forms.py#L35-L43

localflavor/by/forms.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    Mixin made to conform the DRY principle.
+
+    Overrides the to_python methods to return the value.upper()
+    """
+
+    def to_python(self, value):
+        value = super().to_python(value)
+        if value is None:
+            return value
```
benkonrath, line 32: I don't think it's a good idea to use `None` as the default for empty value. `'empty_value': None` should be remove from the `initial_options` and this code should be updated like this: 
```suggestion
        if value in self.empty_values:
            return self.empty_value
```
This will let users set `None` if they prefer this. Otherwise an empty string will be used.

localflavor/by/forms.py
```diff
@@ -0,0 +1,102 @@
...
+
+    Apart from that, it tries to convert all characters to uppercase as
+    the format of belarussian pass number needs that to be right.
+
+    The valid input format is: XX1234567.
+    """
+
+    initial_options = {
+        'regex': r'[A-Z]{2}\d{7}',
+        'max_length': 9,
```
benkonrath, line 60: `max_length` isn't needed as the regex will take care of checking the length. The same applies to other places with `max_length` is set with a regex that enforces the length (below).

tests/test_by/tests.py
```diff
@@ -0,0 +1,136 @@
...
+
+class BYLocalFlavorTests(TestCase):
+
+    def setUp(self):
+        self.form = BYTestForm()
+        self.valid_payload = {
+            '': '',
+        }
+
+    def _gen_pass_id(self, is_valid):
```
benkonrath, line 19: I think it's better to have a few valid and invalid examples like this so that tests will never get a false positive or negative. That said, the chances are pretty low that there will be problems so feel free to keep things like this if this is what you prefer.

benkonrath (changes requested):
Thanks for the PR and sorry for the delay in getting to this.

There are a couple of changes that I think should be made to improve this a little. 
Let me know if you have any comments about the things I mentioned. Or you'd prefer me to just implement the changes (since there was a big delay in the review).

The string freeze for the 3.1 release is planned in two days (20 May). If you have time for the updates in the next two days then we can try to get this in. Otherwise, it can be included in the next release.

benkonrath:
Moved this to the 4.0 milestone as I didn't have time to make the changes in the last couple of days.

PR #428, opened 2021-02-19
Correct US_STATES to sort by state name, as it originally did. #424

### LINKED ISSUE #424: US US_STATES is sorted incorrectly
https://github.com/django/django-localflavor/blob/master/localflavor/us/us_states.py 

The US_STATES enum is sorted by state code, but STATE_CHOICES (and old Django versions such as https://github.com/django/django/commit/4b2b2eda24) are sorted by state name.  The net effect is that Nevada SHOULD sort before New Jersey, but in fact sorts after since NV is after NJ.

ses4j:
Fix #424.


benkonrath:
Thanks!

Issue #432, opened 2021-03-27
Update Chile to have new region


HectorAsencio:
CL: In 2018 the Bio Bio region was divided into two: Bio Bio (unchanged) and XVI Región de Ñuble (new).

benkonrath:
Thanks for letting us know about the update. Do you have time to make a pull request to add this new region? Thanks.

HectorAsencio:
Pull Request: https://github.com/django/django-localflavor/pull/433

benkonrath:
Fixed via https://github.com/django/django-localflavor/commit/ad62cfe384accbdbddfb974b53ed1be76f96276b

PR #433, opened 2021-03-29
Added new Region del Ñuble in CL

HectorAsencio:
Added Región del Ñuble in CL choices. Updated test to match choices list.


HectorAsencio:
PR for #432

benkonrath:
Looks good. Could you add a changelog entry so that people know that this change is coming in when they update? Thanks!

benkonrath:
Thanks @HectorAsencio!

PR #434, opened 2021-04-06
Added russian policy (or OMC) form

Malomalsky:
Added Russian insurance policy (or OMC) form to RU. 

Source: https://en.wikipedia.org/wiki/Federal_Compulsory_Medical_Insurance_Fund_(Russia)


benkonrath:
This needs at least one test before we can add it. Does the OMC have a check digit or anything to verify that the number has been correctly entered? If so, we should validate the number as well.

PR #440, opened 2021-05-16
Added error codes to ValidationErrors

### LINKED ISSUE #338: Propposal: Include `code` in ValidationError
Most of the django core validators like [EmailValidator](https://github.com/django/django/blob/master/django/core/validators.py#L191), set the `code` parameter to `invalid`. I think that this can also be done in localflavor.

For example this can be used in testing to check a validation without need to parse the localized message.

benkonrath:
Fixes #338

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -65,7 +65,7 @@ Other changes:
   `forms.CharField` and `forms.Field`. Positional arguments are not supported in the
   the parent form and did not work
   `gh-421 <https://github.com/django/django-localflavor/pull/421>`_).
-- Add error codes to all `ValidationError`s as recommended by
+- Added error codes to all `ValidationError`s as recommended by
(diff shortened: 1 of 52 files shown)


docs/changelog.rst
```diff
@@ -65,6 +65,9 @@ Other changes:
   `forms.CharField` and `forms.Field`. Positional arguments are not supported in the
   the parent form and did not work
   `gh-421 <https://github.com/django/django-localflavor/pull/421>`_).
+- Add error codes to all `ValidationError`s as recommended by
+  `Django's form validation documentation <https://docs.djangoproject.com/en/3.2/ref/forms/validation/#raising-validationerror>`_
```
claudep, line 69: I think you can replace `3.2` by `stable`.

benkonrath: Thanks, I didn't know about that redirect.

PR #443, opened 2021-06-10
Add FRRNAField

babastienne:
Added FRRNAField to allow the use of RNA French Association Identification Number.
More information on https://fr.wikipedia.org/wiki/R%C3%A9pertoire_national_des_associations or https://www.associations.gouv.fr/le-rna-repertoire-national-des-associations.html

**All Changes**

      already there.

**New Fields Only**

      tax / VAT ID etc.

      localflavors.


benkonrath:
Thanks for the PR. Sorry for the delay in merging it.

PR #444, opened 2021-06-19
Fix spelling of the India state of Chhattisgarh

abhineet97:
According to the official [government website of Chhattisgarh](https://www.cgstate.gov.in/en/web/guest/home), the English spelling contains two 'h' instead of one.

**All Changes**

      already there.


benkonrath:
Thanks!

PR #446, opened 2021-06-23
Inherit from CharField instead of Field

### LINKED ISSUE #445: Field of CharField inheritance
Most of localflavor form fields inherit from `CharField`. However, some are directly inheriting from `Field`. Whenever you use those fields as the default form field for any model CharField, you'll get the `__init__() got an unexpected keyword argument 'max_length'` TypeError. I wonder if the fact some fields inheriting from `Field` only is a desing decision or simply an oversight.

@benkonrath, what do you think?

benkonrath:
A first (incomplete) PR for converting the fields inheriting from `Field` to inheriting from `CharField`.

Fixes  #445

### DIFF
--- a/localflavor/at/forms.py
+++ b/localflavor/at/forms.py
@@ -2,7 +2,7 @@
 import re
 
 from django.core.validators import EMPTY_VALUES
-from django.forms import ValidationError, CharField, RegexField, Select
+from django.forms import CharField, RegexField, Select, ValidationError
 from django.utils.translation import gettext_lazy as _
 
 from .at_states import STATE_CHOICES
--- a/localflavor/at/forms.py
+++ b/localflavor/at/forms.py
@@ -1,8 +1,7 @@
 """AT-specific Form helpers."""
 import re
 
-from django.core.validators import EMPTY_VALUES
-from django.forms import ValidationError, CharField, RegexField, Select
+from django.forms import CharField, RegexField, Select, ValidationError
 from django.utils.translation import gettext_lazy as _
 
 from .at_states import STATE_CHOICES
--- a/localflavor/at/forms.py
+++ b/localflavor/at/forms.py
@@ -2,7 +2,7 @@
 import re
 
 from django.core.validators import EMPTY_VALUES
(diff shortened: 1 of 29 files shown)


localflavor/at/forms.py
```diff
@@ -53,8 +52,8 @@ class ATSocialSecurityNumberField(Field):
 
     def clean(self, value):
         value = super().clean(value)
-        if value in EMPTY_VALUES:
-            return ""
+        if value in self.empty_values:
+            return self.empty_value
```
claudep, line 56: Could/should those empty_values changes be part of a separate commit?

claudep:
I would also make a separate PR for isort changes.

benkonrath:
The isort changes are mostly relevant but I get your point - reverted.

localflavor/at/forms.py
benkonrath: The reason I included these changes was because changing to `CharField` would introduce the option to set `empty_value` as an argument without it having any effect. `self.empty_value` defaults to an empty string and `self.empty_values` defaults to `EMPTY_VALUES` so there shouldn't be a problem with this change.

Does this reasoning sound ok to you? I'm definitely open to less invasive changes if you think that's better for a 3.1.1 release and then following up with the empty value changes in 4.0.

claudep: Less hard-coding sounds fine!

claudep: In fact, I would simply return `value`, as the conversion of the value to self.empty_value is already happening in `to_python`. Even if the result should be exactly the same.

benkonrath: Makes sense. I've been using the `return self.empty_value` pattern for a while and I think it should be changed everywhere because existing fields are often used a reference when new fields are added. This can happen in another PR.

claudep:
The more I think about it, the less I like those `strip` parts. As you mentions, the default is `True` anyway, so I don't see the point in twiddling with that. I tried your PR without those parts in #462 and tests seem to pass (at least locally).

benkonrath:
That's fine. Feel free to merge your modified version.

PR #463, opened 2021-11-02
added model tests for nepal local-flavour

sandeepN06:
@benkonrath as per https://github.com/django/django-localflavor/pull/460#issuecomment-954078190 comment , i have added tests for nepal's local-flavour model.


sandeepN06:
@benkonrath can you review it

claudep:
@sandeepN06 Note that Ben is maintaining this package during his free time, so you should have a little bit more patience before pinging him for a review. I'm sure you understand!

sandeepN06:
My sincere apologies.

tests/test_np/tests.py
```diff
@@ -37,3 +39,45 @@ def test_NPPostalCodeFieldTest(self):
...
+        })
+
+    def test_get_display_methods(self):
+        """Test that the get_*_display() methods are added to the model instances."""
+        place = self.form.save()
+        self.assertEqual(place.get_district_display(), 'Achham')
+        self.assertEqual(place.get_zone_display(), 'Mahakali')
+        self.assertEqual(place.get_province_display(), 'Bagmati')
+
+    def test_errors(self):
```
benkonrath, line 63: I think this test isn't necessary. The postal code form field is already tested and testing the invalid choices in the forms doesn't add extra value because we assume that Django's Select is already tested to limit the choices to what's set in `choices`.

We need tests that use the model fields to ensure they are working as expected. You can also look at what I did with GH. I just set a valid value for `region`, called `clean_fields()` to check that it accepts a valid choice, called `save()` to verify that saving the models works and checked  `get_region_display()` (the method created by Django to get the presented text value) to see if it was correct.

https://github.com/django/django-localflavor/blob/master/tests/test_gh/tests.py#L19

If you do something like this for all of the model fields I think it will be good to go. Thanks.

benkonrath (changes requested):
Thanks for your patience. I've added suggestion how I think the models can be tested. Let me know what you think.

sandeepN06:
Does that fit @benkonrath ?

benkonrath (approved):
Look good! :+1:  Thanks for your help adding these tests.

PR #468, opened 2022-04-21
Removed curp inconvenient word from mx flavor

quevon24:
Removed "VACA" from CURP_INCONVENIENT_WORDS for MX flavor.

Added valid curp to test case.

**All Changes**

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -69,6 +69,7 @@ Modifications to existing flavors:
   - `ro.forms.ROCountyField`
   - `tr.forms.TRIdentificationNumberField`
   - `us.forms.USStateField`
+- Removed inconvenient word from CURP_INCONVENIENT_WORDS for MX flavor
 
 Other changes:
 
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -69,7 +69,7 @@ Modifications to existing flavors:
   - `ro.forms.ROCountyField`
   - `tr.forms.TRIdentificationNumberField`
   - `us.forms.USStateField`
-- Removed inconvenient word from CURP_INCONVENIENT_WORDS for MX flavor
+- Removed inconvenient word VACA from CURP_INCONVENIENT_WORDS for MX flavor
 
 Other changes:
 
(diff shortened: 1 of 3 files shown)


docs/changelog.rst
```diff
@@ -69,6 +69,7 @@ Modifications to existing flavors:
   - `ro.forms.ROCountyField`
   - `tr.forms.TRIdentificationNumberField`
   - `us.forms.USStateField`
+- Removed inconvenient word from CURP_INCONVENIENT_WORDS for MX flavor
```
benkonrath, line 72: It would be nice to specifically mention that VACA was removed.

benkonrath (commented):
Thanks for the PR. Do you have some context about this change? Perhaps an updated URL for CURP_INCONVENIENT_WORDS as the URL in the comment doesn't seem to be working. Thanks.

quevon24:
> Thanks for the PR. Do you have some context about this change? Perhaps an updated URL for CURP_INCONVENIENT_WORDS as the URL in the comment doesn't seem to be working. Thanks.

I was working with some curps, and suddendly one starting with VACA appeared, of course didn't passed the validation, then I opened https://www.gob.mx/curp/ which is the government site where you can check the validity of the CURP, and was valid, then I inspected localflavor and realized that VACA is included on the inconvenient words, then I searched for official documentation, but the one that comes with inconvenient words is quite old, there is no recent documentation.

The most recent is here: https://www.gob.mx/cms/uploads/attachment/file/681698/reglas_para_la_ejecucion_de_los_procedimientos_asignacion_de_la_curp.pdf

But there's no inconvenient words list in there.

The documentation says that when your CURP is generated, if one inconvenient word appears, the first vowel shoud be replaced with an X, so VACA will become VXCA but when you check the CURP on government site is not valid if you write it as VXCA, you need to replace the X with the vowel to work. Also the official document generated there comes without the X.

That's why it comes that VACA is a valid word. Official page accept it.

I'll update the changelog

benkonrath:
Thanks for the info and the update. :+1: If somebody's curious about the change they'll be able to find your explanation here.

Issue #464, opened 2021-11-11
Bug: CIF spanish starting with U does not validate


Vargax01:
Hello,
django-localflavor ES does not validate the cif that begin with U. I have been analyzing the code and I have seen that the U is not within the list of values ​​of the cif.

`self.cif_types = 'ABCDEFGHJKLMNPQRSVW'`

The cif that begin with U are totally valid (https://es.wikipedia.org/wiki/C%C3%B3digo_de_identificaci%C3%B3n_fiscal)

Could you tell me what to do to fix this.

Thanks greetings

benkonrath:
Hi, Thanks for the bug report. The best thing to do is submit a PR with the fix and passing test demonstrates that the issue is fixed. Thanks.

sandeepN06:
Hey @Vargax01 you could fork it and clone this repo in your pc and redirect to ES localflavour and you could alter the cif_types and run the tox if all tests get passed send a PR. If you face any difficulties just put a message

Vargax01:
Hi,
Yes I solve the problem, I edit the code in the forms.py file and change the line to this line: 
`self.cif_types = 'ABCDEFGHJKLMNPQRSUVW'`
And everything works perfectly.
I hope you can implement the updated cif letters in future versions

Thanks greetings

sandeepN06:
> Hi,
> Yes I solve the problem, I edit the code in the forms.py file and change the line to this line: 
> `self.cif_types = 'ABCDEFGHJKLMNPQRSUVW'`
> And everything works perfectly.
> I hope you can implement the updated cif letters in future versions
> 
> Thanks greetings

You need to create a PR to get your code merged

benkonrath:
Fixed in #469

PR #469, opened 2022-04-21
CIF spanish starting with 'U' bug resolved 

sandeepN06:
**Please replace these instructions with a description of your change. The
'New Fields Only' section should be removed if your pull request
doesn't add any new fields.**

Thanks for your contribution!

Issue : (https://github.com/django/django-localflavor/issues/464)

first version.

**All Changes**

      already there.

**New Fields Only**

      tax / VAT ID etc.

      States postal system.

      localflavors.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -23,6 +23,8 @@ New fields for existing flavors:
 
 Modifications to existing flavors:
 
+- CIF spanish starting with 'U' bug resolved
+  (`gh-469 <https://github.com/django/django-localflavor/pull/469>`_).
 - Fix error code for BRPostalCodeValidator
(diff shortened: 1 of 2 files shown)


benkonrath (changes requested):
Do you have a reference for this change? For instance, an official website that states the CIF types? I'm just wondering why this was missed in the first implementation. Did something change?

This PR itself needs a couple of additional things:
* a test with a CIF number that has a U (if possible)
* a changelog entry

Thanks.

sandeepN06:
@benkonrath  This is the reference for the change : https://es.wikipedia.org/wiki/C%C3%B3digo_de_identificaci%C3%B3n_fiscal

benkonrath:
Ok, thanks!

benkonrath (approved):

PR #470, opened 2022-05-09
GitHub Action to lint Python code

cclauss:
`print()` is a function in Python 3.


claudep (changes requested):
There is already a lint target in the current test workflow. I think it would be fair to complete that one with a `flake8` command. I don't think we need all the other linters for now.

PR #477, opened 2022-12-02
Improve German translation

raphaelm:
The current German translation file contains some

- grammatical errors
- typos
- stylistic inconsistencies between formally addressing the user ("Sie") and avoiding it
- inconsistencies within the source

I attempted to resolve these and also add some missing translations in the process.

Sorry for not using Transifex, I hope this can still be merged -- I just found out that it is in use after finishing my work. I couldn't find it linked anywhere in the repo or the docs. (Its probably also possible to just upload my .po file to Transifex. If someone gets less confused by the Transifex web interface than I do whenever I use it, you have my permission to do so.)


claudep:
Thanks, however translation work happens on https://explore.transifex.com/django/django-localflavor/

raphaelm:
Sure, I've uploaded my updated file at Transifex now. Looking forward to the release using it :)

PR #486, opened 2023-04-19
Added check for BBAN in IBAN Validator

### LINKED ISSUE #485: IBAN validator failed for Italy
IBANValidator doesn't check properly iban for Italy. It hasn't raised an ValidationError for below iban:
IT2813815463652787128285355

Missing is Country code:
https://wise.com/hk/iban/italy

benkonrath:
Also:
* Updated IBAN_SEPA_COUNTRIES to latest data.
* Updated IBAN_COUNTRY_CODE_LENGTH to latest data.

I used `stdnum.iban` to do a validation as the last step. I did this so that we can get this into the 4.0 release without string changes. We should probably replace the existing code with calls to the appropriate functions in `stdnum.iban` but I didn't want to do this now because of the string freeze (and I don't have too much time right now). Thoughts?

Fixes #485


tests/test_generic/tests.py
```diff
@@ -164,9 +165,9 @@ def test_iban_fields(self):
             'MU17 BOMM 0101 1010 3030 0200 000M UR': 'MU17BOMM0101101030300200000MUR',
             'MU 17BO MM01011010 3030-02 000-00M UR': 'MU17BOMM0101101030300200000MUR',
 
-            'BE68539007547034': 'BE68539007547034',
-            'BE68 5390 0754 7034': 'BE68539007547034',
-            'BE-685390075470 34': 'BE68539007547034',
+            'BE31538007547055': 'BE31538007547055',
+            'BE31 5380 0754 7055': 'BE31538007547055',
+            'BE-315380075470 55': 'BE31538007547055',
```
benkonrath, line 170: The example BE IBAN that we were using (from the SWIFT IBAN docs) passes the IBAN validation, but not the stricter validation that stdnum.iban uses for some countries. That's the reason I needed to use a different IBAN here.

claudep (approved):
Very nice, thanks!

PR #488, opened 2023-04-21
Removed support for outdated Python 3.6

claudep:


benkonrath:
I re-added tests on Python 3.6 because Django 3.2 does run with Python 3.6 even though Python 3.6 is EOL.

Do you know if the latest Django releases for 3.2 also stop testing on Python 3.6? If Django itself stopped testing on 3.6 in 3.2x, then we should also stop testing on 3.6 (merge this PR). However, if Django 3.2.x still tests on Python 3.6, I think we should too. I did a quick look but I wasn't able to find the github actions config for the 3.2.x branch.

claudep:
Indeed, Django 3.2 is still tested with Python 3.6: https://djangoci.com/job/django-3.2/

So we can delay this PR.

PR #495, opened 2023-09-23
Dropped support for EOL Python 3.6 and 3.7

claudep:


claudep:
@benkonrath, I cannot spot the reason of the prospector failure. Do you have any idea?

benkonrath:
I'm not sure what's going on either. I think one of the pylint dependencies has been updated and is causing a problem. The logs for the previous runs have been deleted so it will be hard to know which dependency to pin to fix this - if it's even the problem.

I did however notice that prospector is working with Python 3.9 on my local machine. I think the easiest fix is to use Python 3.9 in tox as well. Just change python3.11 to python3.9 here and I think it will work again:
https://github.com/django/django-localflavor/blob/c539612758ded5fcb3fbb5cb5ff20f1b0de96fcc/tox.ini#L47

Side note: We're using a pretty old version of prospector and it could be updated as a maintenance task at some point. I haven't done the update yet because prospector continually adds new rules and therefore finds more "problems" in the code. Some of these problems are legitimate, but some are more on the opinion side of the fence. The end result of course is that an update takes more work than just bumping the numbers.

I suppose an update to prospector could involve disabling some of checks to keep things manageable. But this is probably better for a separate topic. Hopefully the quick fix I mentioned will get this PR going.

benkonrath:
The PR looks good by the way (besides the CI failure which seems unrelated). Thanks!

claudep:
Thanks for your help. I'll merge this one.

PR #497, opened 2023-09-23
Fixed prospector lint CI job

benkonrath:


claudep:
Looks good, thanks! If recent prospector/pylint are too invasive, maybe we should simply go back to some simpler linter (flake8 or ruff). I'm not a fan of black, by the way.

benkonrath:
I'm open to something simpler, however, I think other linters may have the same issue as prospector; upgrading always will add more rules and therefore take time. Perhaps prospector is a little more complex because it depends on so many of the Python linters.

I do actually like black but I see it more as a code formatter rather than a static analysis linter that tries find potential problems in the code. My argument for black is that it's nice not to waste time commenting on formatting in reviews or thinking about it when coding. When I first started using black, I didn't always like the style of code it produced but now I'm used to it and I don't really notice anymore.

The negative point for using black is that we'd loose git history if we ever did start using it. This is something I'd miss. In any case, I think all active maintainers of localflavor would need to be onboard for a change like this so we can just keep things the way they are (without black).

I'll revisit updating prospector or switching to flake8 when I have a little more spare time.

PR #501, opened 2023-10-20
Update django.po

mikkelpetersen:
Updated translation for Danish.

**All Changes**

- [ ] Add an entry for your name in the docs/authors.rst file if it's not
      already there.


claudep:
Please, update the translations on Transifex directly: https://explore.transifex.com/django/django-localflavor/

Issue #502, opened 2023-11-26
Update Python and Django versions


pauloxnet:

pauloxnet:
Solved in #503

PR #503, opened 2023-11-26
Fix # 502 -- Update Python and Django versions

pauloxnet:
**All Changes**

      already there.

Ref #502


claudep (approved):
Thanks Paolo!

pauloxnet:
> Thanks Paolo!

Thanks for approving so fast.

Do you think we need to have a release before Django 5.0 will be published?

Issue #504, opened 2024-03-27
Accept French Postal Services identifiers in forms


Chatewgne:
The form validation in `FRSIRENENumberMixin`  is missing a special case from the SIRENE register.

The identifiers for Postal Services need a special validation process and are not currently validated.

Source (in french) : [https://www.sirene.fr/static-resources/doc/lettres/lettre-16-novembre-2013.pdf](https://www.sirene.fr/static-resources/doc/lettres/lettre-16-novembre-2013.pdf)

benkonrath:
Fixed in #505.

PR #505, opened 2024-03-27
fix: Accept French Postal Services identifiers in forms

Chatewgne:
This PR is intended to fix the following issue : https://github.com/django/django-localflavor/issues/504

The form validation in `FRSIRENENumberMixin`  is missing a special case from the SIRENE register.

The identifiers for Postal Services need a special validation process and are not currently validated.

Source (in french) : [https://www.sirene.fr/static-resources/doc/lettres/lettre-16-novembre-2013.pdf](https://www.sirene.fr/static-resources/doc/lettres/lettre-16-novembre-2013.pdf)

**All Changes**

      already there.

### DIFF
--- a/docs/authors.rst
+++ b/docs/authors.rst
@@ -33,6 +33,7 @@ Authors
 * Ben Konrath
 * Bruno M. Custódio
 * Burhan Khalid
+* Célia Prat
 * Claude Paroz
 * Daniel Ampuero
 * Daniela Ponader
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -17,6 +17,8 @@ Modifications to existing flavors:
 - Fix Belarus passport field description punctuation
   (`gh-484 <https://github.com/django/django-localflavor/pull/484>`_).
 - Change `Kiev` to `Kyiv` 🇺🇦 according to ISO_3166-2:UA
+- Accept French Postal Services identifiers in forms
+  (`gh-505 <https://github.com/django/django-localflavor/pull/505>`_).
 
 Other changes:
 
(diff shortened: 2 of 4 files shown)


claudep:
Sorry for the delay for reviewing, could you please add/complement a test for this?

Chatewgne:
Thanks for getting back to me @claudep, I have added the test cases :)

benkonrath:
The exception only applies to SIRET, not to SIREN, so this update isn't correct. I think we should remove the mixin and use stdnum to validate both of the fields directly in the form class. This would add few lines of common code to both classes but it makes things clearer.

claudep:
Any reason to turn this PR as a draft?

Chatewgne:
> The exception only applies to SIRET, not to SIREN, so this update isn't correct. I think we should remove the mixin and use stdnum to validate both of the fields directly in the form class. This would add few lines of common code to both classes but it makes things clearer.

I have removed the mixin as suggested and revised validation for both SIRET and SIREN.

Reminder : the SIREN is included in the SIRET (9 first characters) 

Validation :
- SIREN has 9 chars AND validates luhn algorithm
- SIRET has 14 chars AND validates luhn algorithm OR starts with "356000000" and the sum of the 14 characters is a multiple of 5 

Sources : 
- [https://fr.wikipedia.org/wikiSyst%C3%A8me_d%27identification_du_r%C3%A9pertoire_des_%C3%A9tablissements](https://fr.wikipedia.org/wiki/Syst%C3%A8me_d%27identification_du_r%C3%A9pertoire_des_%C3%A9tablissements)
- [https://www.sirene.fr/static-resources/doc/lettres/lettre-16-novembre-2013.pdf](https://www.sirene.fr/static-resources/doc/lettres/lettre-16-novembre-2013.pdf)
- [https://portal.hardis-group.com/doccenter/pages/viewpage.action?pageId=120357227](https://portal.hardis-group.com/doccenter/pages/viewpage.action?pageId=120357227)

benkonrath:
Thanks!

PR #507, opened 2024-05-15
Added MACinNumberField to MA flavor

AbdelYsf:
Adding CIN field to morocco's falvor, The CIN represents the ID of a Moroccan citizen
It is an 8-max-length string that starts with one or two Latin letters followed by digits,
           with the first digit not being zero.
Implemented as in the official government site "https://www.cnie.ma/"

### DIFF
--- a/localflavor/ma/forms.py
+++ b/localflavor/ma/forms.py
@@ -84,7 +84,7 @@ class MACinNumberField(RegexField):
            with the first digit not being zero.
 
          - as implemented in the official government site "https://www.cnie.ma/"
-
+        .. versionadded:: 4.1
     """
 
     default_error_messages = {
--- a/localflavor/ma/forms.py
+++ b/localflavor/ma/forms.py
@@ -90,10 +90,10 @@ class MACinNumberField(RegexField):
     default_error_messages = {
         'invalid': _('Enter a valid Moroccan CIN number.'),
     }
-    cin_pattern = r'^[A-Za-z]{1,2}[1-9]\d{0,6}$'
+    cin_pattern = r'^[A-Za-z]{1,2}[1-9][0-9]{0,6}$'
 
     def __init__(self, **kwargs):
-        kwargs.setdefault('label', _('Cin Number'))
+        kwargs.setdefault('label', _('CIN Number'))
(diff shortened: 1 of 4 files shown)


localflavor/ma/forms.py
```diff
@@ -74,3 +74,26 @@ class MARegionField(CharField):
...
+        .. versionadded:: 4.1
+    """
+
+    default_error_messages = {
+        'invalid': _('Enter a valid Moroccan CIN number.'),
+    }
+    cin_pattern = r'^[A-Za-z]{1,2}[1-9]\d{0,6}$'
+
+    def __init__(self, **kwargs):
+        kwargs.setdefault('label', _('Cin Number'))
```
benkonrath, line 96: This will be seen by users so I think it should be "CIN Number".
```suggestion
        kwargs.setdefault('label', _('CIN Number'))
```

localflavor/ma/forms.py
```diff
@@ -74,3 +74,26 @@ class MARegionField(CharField):
...
+           with the first digit not being zero.
+
+         - as implemented in the official government site "https://www.cnie.ma/"
+        .. versionadded:: 4.1
+    """
+
+    default_error_messages = {
+        'invalid': _('Enter a valid Moroccan CIN number.'),
+    }
+    cin_pattern = r'^[A-Za-z]{1,2}[1-9]\d{0,6}$'
```
benkonrath, line 93: I think \d in python3 include unicode digits. Try this in a python console:

```
import re; re.match(r'\d', '٠١٢٣٤٥٦٧٨٩')
```

I think it should be this:
```suggestion
    cin_pattern = r'^[A-Za-z]{1,2}[1-9][0-9]{0,6}$'
```

benkonrath (approved):
Thanks, it looks good. Just the two small points address. Let me know if you have any questions or remarks.

localflavor/ma/forms.py
AbdelYsf: Thanks for pointing that out ! , I will update the code accordingly

benkonrath (approved):
Looks good. Thanks!

PR #508, opened 2024-06-08
Kyiv - not Kiev 🇺🇦 

DmytroLitvinov:
**All Changes**
- [X] Change `Kiev` to `Kyiv` 🇺🇦 according to ISO_3166-2:UA. More context about [a problem](https://en.wikipedia.org/wiki/KyivNotKiev)

      already there.


claudep:
Thanks for the fix!

PR #509, opened 2024-06-17
Update no_municipalities.py to reflect recent changes to norwegian counties

hakosl:
Update list of Norwegian counties(fylker) according to changes at the start of 2024: https://en.wikipedia.org/wiki/Counties_of_Norway#List_of_counties

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -48,6 +48,7 @@ Modifications to existing flavors:
   (`gh-490 <https://github.com/django/django-localflavor/pull/490>`_).
 - Update list of Latvia municipalities after municipal reform from 2021
   (`gh-519 <https://github.com/django/django-localflavor/pull/519>`_).
+- Update list of Norwegian counties(fylker) `according to changes at the start of 2024 <https://en.wikipedia.org/wiki/Counties_of_Norway#List_of_counties>`_
 
 Other changes:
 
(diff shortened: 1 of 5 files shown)


docs/changelog.rst
```diff
@@ -17,6 +17,7 @@ Modifications to existing flavors:
 - Fix Belarus passport field description punctuation
   (`gh-484 <https://github.com/django/django-localflavor/pull/484>`_).
 - Change `Kiev` to `Kyiv` 🇺🇦 according to ISO_3166-2:UA
+- Update list of Norwegian counties(fylker) `according to changes at the start of 2024 <https://en.wikipedia.org/wiki/Counties_of_Norway#List_of_counties>`_ (`gh-509 <https://github.com/django/django-localflavor/pull/509/files>`_)
```
claudep, line 20: I think I would mention in release notes the county values that were deleted from the choice tuple, so any dev using the list in a model field is alerted that they probably have to make some data migration.

hakosl: Good idea, i have updated the changelog to also mention what counties are changed.

hakosl:
I fixed the tests

benkonrath:
As a follow up to Claude's comment about the changelog, maybe we should use the same procedure as we did with the 3.1 release when we had breaking data changes.

1. Add a warning to NOMunicipalitySelect.

```
import warnings

class NOMunicipalitySelect(Select):
    def __init__(self, attrs=None):
        warnings.warn("Choices have changed for NOMunicipalitySelect in localflavor 4.1. See changelog for details.")
        super().__init__(attrs, choices=MUNICIPALITY_CHOICES)
```


2. Be very clear about the breaking change at the start of the changelog and provide instructions. The 3.1 release notes are a good example:

https://github.com/django/django-localflavor/blob/master/docs/changelog.rst#31---2021-05-28

@hakosl What do you think?

PR #514, opened 2024-12-24
Switch from Python 2 to 3 in intersphinx_mapping setting

benkonrath:


claudep (approved):
Oups, good catch :+1:

benkonrath:
I just happened to notice it in the tox output. :-)

PR #515, opened 2025-02-17
Fix the Romanian Flavor ROCNP Field

tudoramariei:
- years ending with 00 (2000, 1900, etc.) would appear as invalid
- use .datetime.strptime instead of .date since it's more reliable
- remove an unused import statement


claudep (approved):
Thanks!

PR #516, opened 2025-03-22
Updated Python/Django minimal versions

claudep:


benkonrath:
Look good! :+1:

benkonrath (approved):

claudep:
Thanks Ben! I guess that I should add 5.2 now, and then we might produce a new release.

claudep:
In addition to the previous version, I added support for Django 5.2, removed testing on EOL Python 3.8, and limit testing to minimal/maximal Python version for each Django version. I started doing that recently in some projects, first because I never saw failures with intermediate Python versions, and second as a move to show we do our part to limit CO2 emissions.

benkonrath:
Only testing min/max Python versions seems reasonable. Feel free to merge this.

I'm wondering if we should do a 5.0 given we've stopped testing on older versions of Python and Django. I don't think there are any incompatible changes though. Thoughts?

I can do the release. I'll figure out a reasonable string freeze date and plan a release date - I want to clean up and merge a couple of PRs for this release. I'll have some time for this in a couple of days.

claudep:
Yes, it might make sense to switch to 5.0. It's great if you can spend some time on the package. It's so precious to have people like you committing for long-term maintainership. :heart_eyes:

claudep:
By the way, it would be nice to switch from `setup.py` to `pyproject.toml`. I could find some time to explore that, if you want.

PR #519, opened 2025-04-22
Update Latvia municipalities after municipal reform from 2021

quadracik:
**All Changes**

- [x] Changes to the localflavor/lv/lv_choices.py.
Sources:
https://en.wikipedia.org/wiki/Administrative_divisions_of_Latvia
https://en.wikipedia.org/wiki/ISO_3166-2:LV

      already there.


benkonrath:
Thanks for the update! I added a warning so existing users know that the data has changed, and a notice in the changelog.

PR #529, opened 2025-07-03
Fix REGON validation

Serp-ent:
- Fix regon length regex
- Fix regon validation for 14 digits

Fixed validation, because regex currently allows for strings from 9 to 14 digits, but only fixed 9 digits or 14 digits should be allowed. Additionally the checksum validation for last digit case (where modulo of checksum is equal 10) is handled incorrectly for 14 digits length string.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -14,7 +14,8 @@ New fields for existing flavors:
 
 Modifications to existing flavors:
 
-- None
+- Fix REGON validation for PL
+  (`gh-529 <https://github.com/django/django-localflavor/pull/529>`_).
 
 Other changes:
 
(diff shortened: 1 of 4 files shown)


docs/changelog.rst
```diff
@@ -101,6 +101,8 @@ Modifications to existing flavors:
   (`gh-506 <https://github.com/django/django-localflavor/pull/506/files>`_).
 - Allow formatted IBANs longer than 26 chars to be saved in `IBANField`
   (`gh-522 <https://github.com/django/django-localflavor/pull/522>`_).
+- Fix REGON validation for PL
+  (`gh-529 <https://github.com/django/django-localflavor/pull/529>`_).
```
claudep, line 105: 5.0 has been released, please add your note under the 5.1 section.

claudep (approved):
Thanks, looks good!

@benkonrath, do you want to have a quick look before merging?

benkonrath (approved):
Looks good. Thanks!

PR #530, opened 2025-07-03
feat(tw): init tw with administrative division select

mattwang44:
- Add new flavor for Taiwan (`tw`)
- Add `TWAdministrativeDivisionSelect`

**All Changes**

      already there.

**New Fields Only**

      tax / VAT ID etc.

      States postal system.

      localflavors.

### DIFF
--- a/localflavor/tw/tw_choices.py
+++ b/localflavor/tw/tw_choices.py
@@ -1,0 +1,28 @@
+#: An alphabetical list of cities and counties for use as `choices` in a formfield.
+#: Based on the administrative divisions of Taiwan.
+#: https://en.wikipedia.org/wiki/List_of_administrative_divisions_of_Taiwan
+
+TW_ADMINISTRATIVE_DIVISION_CHOICES = (
+    ("changhua_county", "彰化縣"),
+    ("chiayi_city", "嘉義市"),
+    ("chiayi_county", "嘉義縣"),
+    ("hsinchu_city", "新竹市"),
+    ("hsinchu_county", "新竹縣"),
+    ("hualien_county", "花蓮縣"),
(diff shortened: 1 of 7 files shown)


benkonrath:
Thanks for the PR! I'll have a look at it next week sometime.

localflavor/tw/tw_choices.py
```diff
@@ -0,0 +1,28 @@
+#: An alphabetical list of cities and counties for use as `choices` in a formfield.
+#: Based on the administrative divisions of Taiwan.
+#: https://en.wikipedia.org/wiki/List_of_administrative_divisions_of_Taiwan
+
+TW_ADMINISTRATIVE_DIVISION_CHOICES = (
+    ("changhua_county", "彰化縣"),
```
benkonrath, line 6: The displayed text should be in English and marked as translatable with gettext. We use Transifex to manage the translations - the translated version can be added to Transifix after this PR is merged. The translations are then pulled into the repo before a release, or as needed.

You can see an example of how to do this in the KW local flavor:

https://github.com/django/django-localflavor/blob/master/localflavor/kw/kw_areas.py  

I realize this is a roundabout way of getting the administrative names displayed, but this is how we allow developers to create multi-lingual applications.

benkonrath (changes requested):
Sorry for the delay. The names need to be updated as indicated before this can be merged.

localflavor/tw/tw_choices.py
mattwang44: fixed, please review again.
how can i help with the translation on the Transifex?

benkonrath:
Thanks for PR and for your quick update!

PR #547, opened 2026-04-07
Require stdnum>=2.2 and add example IBANs from IBAN registry v101

benkonrath:
We need to require stdnum >= 2.2 so that all the IBANs from registry v101 validated correctly.

### DIFF
--- a/docs/changelog.rst
+++ b/docs/changelog.rst
@@ -6,8 +6,6 @@ Changelog
 
 New flavors:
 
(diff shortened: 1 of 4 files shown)


benkonrath:
python-stdnum was bumped to 2.0 because Python 2 was dropped. Requiring stdnum >= 2.2 shouldn't be a problem for our users since we only support Python 3.

https://github.com/arthurdejong/python-stdnum/releases/tag/2.0

PR #548, opened 2026-06-23
Fix Python gh-actions test failure

benkonrath:


claudep (approved):
Thanks for fixing this!

